PR #1, opened 2025-11-06
MDS-0 prettier: update config and re-trigger

andrewgos:
## What was done

- Updated configs
  - Ignore generated tokens
  - Set quotes to be single quotes
  - Add import organiser
- Trigger the prettier on the whole codebase


Chocolate-lightning (approved):
Thanks alot Andrew.

The modifications to prettier config look reasonable for us to run with for the time being.

PR #3, opened 2025-11-10
MDS-235 button: implement react-boostrap as starter

andrewgos:
## What was done
This is a starter component code with `react-bootstrap` Button component, further styling and variation to be implemented.
- `Button.tsx` is kept minimal as a wrapper to `react-boostrap`'s
- `.stories.js` is updated, but left as `js` (@Chocolate-lightning is converting it to `ts` on his branch)
- Package `sass` is installed to get SCSS imports working
- Package `bootstrap` is required for its styling, as `react-bootstrap` doesn't come with the styling
- Package `bootstrap` & `react-bootstrap` is installed as `peerDependencies` with the objective that this will be published as a library package

## Dev Decisions
- Bootstrap's CSS is loaded within the scoped CSS `button.scss` -> this allows each component to be modular in case certain components wanna go away from Bootstrap
  - Alternative 1: loading it on a centralized entrypoint `components/index.ts` -> but require restrictions so that users import only from that single entrypoint
  - Alternative 2: only load it on `.storybook/preview.ts` -> but this will require additional steps for users to load it manually when they use it


Chocolate-lightning (approved):
Kia ora @andrewgos,

Thanks for adding in React Bootstrap and tweaking the existing files to make use of it.
I believe this is fine to merge into main.

PR #5, opened 2025-11-11
MDS-197 cicd: Configure coverage tests and more

Chocolate-lightning:
Added dependabot config for updating packages & storybook addons
Added example unit test file for when we add in testing later in sprint
Added GHA file to run storybook testing suite which emcompasses interaction / a11y & coverage tests

Updated StoryBook to ver 10.x
Updated package.json content to point to correct repository after the rename


package.json
```diff
@@ -1,47 +1,70 @@
 {
-  "name": "moodle_design-system",
+  "name": "design-system",
   "version": "1.0.0",
```
andrewgos: Let's change this to 0.0.1 to start with? for the publishing with npm later

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+    -   package-ecosystem: npm
```
andrewgos: the indentation seems off here, yml files should be having 2 spaces as well.

can you help to rerun the prettier? perhaps the command needs to be updated so we dont use a parser

`"format": "prettier --write .`

.github/workflows/test-storybook.yml
```diff
@@ -0,0 +1,28 @@
+name: Storybook & Accessibility test
+#on: deployment_status
```
andrewgos: lets remove this if its unused?

.github/workflows/test-unit.yml
```diff
@@ -0,0 +1,21 @@
+name: JSUnit with coverage
+#on: deployment_status
```
andrewgos: lets remove this if its unused?

package.json
```diff
@@ -1,47 +1,70 @@
...
   "description": "The Moodle Design System",
   "main": "index.js",
   "type": "module",
+  "dependencies": {
+    "react": "^18.3.1",
```
andrewgos: I think we should avoid putting into the `dependencies` if we are publishing it as a package library. We can keep it under `peerDepedencies` as lib requirements for the users, because its not the job of design system to carry react. And we can keep it in `devDependencies` since we need it for storybooks

CMIIW

andrewgos (commented):
Thanks for setting up all these! Overall LGTM

Just a few minor items I'd like to point out early. Noted on the inline comments.

Chocolate-lightning:
Kia ora @andrewgos,

Thanks for taking a look, I have updated the branch and also opened up the prettier which added more changes after your review.

Cheers,

andrewgos (approved):
Thanks for actioning those and the extra effort in cleaning up the ts/vite configs! LGTM 👍

PR #18, opened 2025-11-20
MDS-239 dependabot: Group updates to reduce PR spam

Chocolate-lightning:
Added grouping to catch updates into 3 different pull requests rather than opening new pull requests for all npm dependency updates besides @storybook updates.

Also ran a format which picked up the vitest file. Might look at adding Husky to add pre-commit linting & formating into the repo.


andrewgos (approved):
LGTM 👍 Feel free to merge whenever you're ready

PR #20, opened 2025-11-21
MDS-263 tokens: improve tokens conversion to a more usable state

andrewgos:
## What
- Revamp Style Dictionary build script with custom formatter
  - To treat primitives differently with `--mds` prefix
  - To handle semantic pointing to primitives
  - To handle semantic pointing to another semantic (multi-layer semantics)
  - To handle semantic containing flat value
  - To handle semantic containing bracketed info `(...)` in the token names
  - To simplify the names of the generated output file
  - To skip `description` that may came along in the tokens with invalid format that could break the built-in formatter
- Relocate tokens in `/assets/src` to `/tokens/dtcg` to remove directory abstraction for better clarity


components/src/button/button.css
Chocolate-lightning, line 1: Should we remove button.scss since we are now using plain CSS?

Chocolate-lightning (changes requested):
Overall this looks really solid, just one minor inline comment to look at.

Great to see the token script taking a more finalised form and adding those transform groups.

Cheers,

Chocolate-lightning:
FYI: The readme should be updated to no longer mention SCSS and reflect the new state of the conversion script.

components/src/button/button.css
andrewgos: ah good pickup! missed that

andrewgos:
Feedback actioned:
- Removed the SCSS file
- Changed SCSS to CSS in the README
- Improved variable naming within `token.ts` script and added comments for those non-self-explanatory conditions

Also, FYI @Chocolate-lightning I've made the loading of bootstrap on top of the component css file `components/src/button/button.css` instead of the css aggregator file, cos I figured that we may want to leave the option open to opt out some components away from react-bootstrap.

Chocolate-lightning (approved):
Thanks for actioning my suggestions. It now looks good and the script is still running well locally.

PR #25, opened 2025-11-21
MDS-250 npm: Add Vite build & GHA

Chocolate-lightning:
Additions:
- Export the Button (for now) via index.tsx within the Design System
- Add & configure Vite builds of the Design System to /dist
- Add GHA to build & publish the Design System

Updates:
- Remove megahack in Storybook GHA as it seems to pass naturally after recent changes
- Modify Button styles file to demonstrate token usage 
- Update vitest.config.ts after running npm run format (Argument to be made for rolling in Husky)

### DIFF
--- a/.github/workflows/test-storybook.yml
+++ b/.github/workflows/test-storybook.yml
@@ -13,10 +13,12 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v6
         with:
-          node-version: 22.12.0
+          node-version: 22.13.0
 
-      - name: Install dependencies
-        run: npm ci
+      - name: Clean install
+        run: |
+          rm -rf node_modules package-lock.json
+          npm install
 
       - name: Run tests
         run: npm run test-storybook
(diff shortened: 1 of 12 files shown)


.github/workflows/test-storybook.yml
```diff
@@ -13,15 +13,12 @@ jobs:
...
         uses: actions/setup-node@v6
         with:
-          node-version: 22.12.0
+          node-version: 22.13.0
 
-      - name: Install dependencies
-        run: npm ci
+      - name: Clean install
+        run: |
+          rm -rf node_modules package-lock.json
```
andrewgos, line 20: why the package-lock.json needs to be cleared also? 🤔

andrewgos (approved):
nicely done!

just a question on something I dont quite understand, I noted it on the exact line using the inline comment

.github/workflows/test-storybook.yml
Chocolate-lightning: It was something suggested to help ensure that the install is absolutely clean after getting the playwright image.

PR #29, opened 2025-11-24
MDS-263 tokens: convert px to rem and append suffixes

andrewgos:
## What
- Added custom transform: css/dimension-px-to-rem

## Why
- Built-in transform needs the type to be 'dimension', but our tokens solely use 'number' type.
- The exported values in DTCG JSON from ZeroHeight would not have the units (ie. 1rem) for dimension-related tokens


Chocolate-lightning (approved):
Kia ora @andrewgos,

Great work as always,

Ignoring the resulting files and focusing on the transformation script, I see that the changes are sensible considering the limitations we are hitting with our token export i.e. not getting dimension typings and having to create a custom parser.

I do wonder if having the pre-commit trigger a token build is required or if we could look at after hooks we could leverage in Husky but that is not really important at the moment and something we could delve into as the project becomes more stable.

Feel free to merge this in.

PR #34, opened 2025-11-25
MDS-263 button: apply the tokens to the component

andrewgos:
## What
Apply latest tokens to the button components as theming on top of react bootstrap


components/src/button/Button.stories.tsx
```diff
@@ -28,11 +28,55 @@ const meta = {
   },
   tags: ['autodocs', 'test', 'stable'],
   // More on argTypes: https://storybook.js.org/docs/api/argtypes
-  argTypes: {},
+  argTypes: {
+    label: {
+      description: 'Button label.',
+      defaultValue: { summary: '' },
```
Chocolate-lightning: Getting the following notice in my IDE:

Deprecated symbol used, consult docs for better alternative

components/src/button/Button.stories.tsx
```diff
@@ -28,11 +28,55 @@ const meta = {
...
+    },
+    variant: {
+      control: { type: 'select' },
+      options: [
+        'primary',
+        'secondary',
+        'danger',
+        'outline-primary',
+        'outline-secondary',
+        'outline-danger',
```
Chocolate-lightning, line 46: Should we add stories for thee outline variants so that our testing covers them?

components/src/button/button.css
```diff
@@ -2,5 +2,135 @@
...
+  line-height: var(--line-height-paragraph-small);
+  letter-spacing: var(--letter-spacing-default);
+}
+.mds-btn.btn:hover {
+  background-color: var(--state-primary-hover);
+}
+.mds-btn.btn:active {
+  background-color: var(--state-primary-active);
+}
+.mds-btn.btn:disabled {
```
Chocolate-lightning, line 23: Is there a primary disabled token available for the border? or is the current implementation intended?

components/src/button/Button.stories.tsx
```diff
@@ -28,11 +28,55 @@ const meta = {
...
+    size: {
+      control: { type: 'select' },
+      options: [undefined, 'sm', 'lg'],
+      description: 'Button size. Default is "md" if not set.',
+      table: {
+        type: { summary: 'sm | lg' },
+        defaultValue: { summary: 'undefined' },
+      },
+    },
+    disabled: {
```
Chocolate-lightning, line 66: We should have at least one story covering the disabled state.

components/src/button/Button.tsx
```diff
@@ -7,6 +7,15 @@ import './button.css';
 
 export interface ButtonProps extends RBButtonProps {
```
Chocolate-lightning, line 8: Is it worth moving away from the dummy test cases I added into Button.test.tsx and making some proper JS Unit tests so that we can get some meaningful coverage?

components/src/button/button.css
```diff
@@ -2,5 +2,135 @@
...
+  background-color: var(--state-secondary);
+  color: var(--body-text-emphasis);
+}
+.mds-btn.btn.btn-secondary:hover {
+  background-color: var(--state-secondary-hover);
+}
+.mds-btn.btn.btn-secondary:active {
+  background-color: var(--state-secondary-active);
+}
+.mds-btn.btn.btn-secondary:disabled {
```
Chocolate-lightning, line 40: Why is there an addition class of btn being added between  .mds-btn & .btn-secondary:disabled

This occurs in other types as well.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Overall, the changes look good. I have added some questions and action items inline to review before we look at merging in this feature.

Cheers,

components/src/button/button.css
andrewgos: the border while disabled wasn't intended, the translucent border attribute wasnt propagating properly to the disabled state. fixed that by increasing css specificity

components/src/button/button.css
andrewgos: It was to increase the specificity of the CSS to override what the Bootstrap has.

It was a solution for a quick turnaround. Let me know if you have other better way!

components/src/button/Button.stories.tsx
andrewgos: fixed! let me know if the notice is still showing up on your IDE

components/src/button/Button.stories.tsx
andrewgos: ah forgotten about those. thanks! added now

components/src/button/Button.stories.tsx
andrewgos: added 👍

components/src/button/Button.tsx
andrewgos: sorry i didnt quite understand which "proper JS Unit tests" that you are referring to, isnt the Button.test.tsx is the unit test?
I forgot that we have `Button.test.tsx`now , will update the test cases there

Chocolate-lightning: No worries, I was thinking if we should start updating the unit tests now that we are moving in a more clear direction rather than only having our interaction & accessibility tests.

Although it'd likely not have much in there and only confirm that props come through or not since its such a low level component.

Not a deal breaker and something we can look at down the line.

andrewgos: Seeing the timeline of steel thread and the situation where we still need to figure out the best "tests" we need for this design system repo, lets settle this on a separate PR.

PR #56, opened 2025-12-04
feat(button): Improve variant restriction and tidy up unit tests

andrewgos:
## JIRA Ticket
https://moodle.atlassian.net/browse/MDS-345

## Ideology behind the change
- Unit test and fuzz tests shouldn't need to depend on stories from storybook
    - Decoupling it from stories will make it more manageable as we scale
    - `npm run test-storybook` already handles smoke testing all stories
- Unit test should worry/focus on the component logic that we added/customised, instead of testing the whole functionality of react-boostrap

### DIFF
--- a/components/src/button/Button.test.tsx
+++ b/components/src/button/Button.test.tsx
@@ -1,135 +1,51 @@
-import { composeStories } from '@storybook/react';
-import '@testing-library/jest-dom';
 import { render, screen } from '@testing-library/react';
 import fc from 'fast-check';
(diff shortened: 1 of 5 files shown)


Chocolate-lightning (approved):
To provide some context, The usage of stories within tests was in line with https://storybook.js.org/docs/writing-tests/integrations/stories-in-unit-tests but we can run with a simpler approach until we see indications otherwise.

The simplification of the typing is also nice to see.

PR #63, opened 2025-12-05
fix(GHA): Add id-token permission to the StoryBook deploy

Chocolate-lightning:
Add missing permission so StoryBook can deploy to GitHub Pages.


andrewgos (approved):
🚀

PR #69, opened 2025-12-09
fix: Add and relocate community insight files

Chocolate-lightning:
Whats changed?

Added: Code of Conduct
Added: Pull request template (No more empty PR messages!)
Changed: Relocated some existing community files into .github
Changed: Sneakily move our linting files into .github as they are typically invoked on commits and GHA


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+## Screenshots/Previews
+
+<!-- If your changes affect the UI, reference Chromatic visual diffs, add screenshots, or link to Storybook previews as needed. -->
+
+## Checklist
+
+- [ ] I have updated, added, and run tests such as JSUnit, Storybook interactions, or fuzzing to prove my fix is effective or that my feature works
+- [ ] I have updated or added Storybook stories for new or changed components (if appropriate)
+- [ ] I have considered accessibility and described any improvements or issues
+- [ ] I have followed [CODEOWNERS](./CODEOWNERS) and requested appropriate reviewers
```
andrewgos: Codeowners should be automatically requested by Github itself, so I reckon this checklist is not needed. I can confirm that I'm receiving the notification by this PR you created

Screenshot 2025-12-09 at 3 10 21 pm


FYI Alternatively, if you need to create a PR without requesting for review, then you can use "Create draft pull request"
Screenshot 2025-12-09 at 3 10 07 pm

.github/linters/.commitlintrc.ts
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   ignores: [
-    (commit) => /^chore(\(main\))?: release \d+\.\d+\.\d+/.test(commit),
+    function (commit) {
```
andrewgos: looks identical to me 👀  why not the arrow function?

.github/linters/.prettierrc
andrewgos, line 1: unfortunately for `.prettierxxx` and `.editorconfig` might need to stay in the root.. otherwise it wouldn't work natively.

I tested it out in my IDE, on save, it no longer detects the configuration set on the `.prettierrc`, and the prettier works on files previously ignored as noted in `.prettierignore`. And `.editorconfig` also not loading properly on my IDE with the tab size of 2, doing tab of 4 on .md files

Plus, with this, it became a crucial requirement to run it by `npm run format` instead of running it plainly without config `npx prettier --write .` which might be overlooked by other devs

andrewgos (changes requested):

.github/linters/.commitlintrc.ts
Chocolate-lightning: This appears to just be an artefact from running linting and formatting in the branch. Will be reverted.

.github/PULL_REQUEST_TEMPLATE.md
Chocolate-lightning: Great call, removed from the template.

.github/linters/.prettierrc
Chocolate-lightning: Fair play on the editorconfig, JetBrains IDEs can pick it up but I suppose we will leave it in the root.

With Prettier, I have updated the vscode settings to direct it to the location. Hopefully that will resolve the issue and we can retain the new location.

andrewgos: cool2 i tested the VSCode settings and it works. we just need to add for the ignore file too
`"prettier.ignorePath": ".github/linters/.prettierignore"`

andrewgos (approved):
Thanks for supporting the vscode settings. Good to go 🚀

PR #103, opened 2026-02-02
feat(tokens): First release of tokens for standalone consumptions

andrewgos:
## Description

What was done in the code:
* Prefixes are now applied to both primitives and semantic tokens
* No minification happens on the output build
* Created an entrypoint file (`/index.ts`) for vite lib build that does the entire loading of the styled design system in order
* Rearranged CSS loading to get clearer the separation between JS & CSS within `components/` and `tokens/`
* Removed redundant `components/src/` folder layer
* Style Dictionary will now be using the built-in formatter instead of the custom formatter
* SCSS files are generated under `tokens/scss`

Reasoning for these changes can be found in the JIRA ticket 

Exposed entrypoints:
- `import { Button } from '@moodlehq/design-system';` -> import the components
- `import '@moodlehq/design-system/css';` -> import the compiled styling (bootstrap, tokens, components styling)
- `import '@moodlehq/design-system/tokens/css';` -> import css tokens only
- `import '@moodlehq/design-system/tokens/scss';` -> import scss tokens only

https://moodle.atlassian.net/browse/MDS-432

## Type of Change

- [x] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Screenshots/Previews
* installed npm package in LMS
npm installed into LMS

* Copied into `public/theme/boost/scss`
copied into public:theme:boost:scss

* Updated `public/theme/boost/scss/preset/default.scss`
updated default scss

* Running `grunt scss` resulted a change in `public/theme/boost/style/moodle.css`
after grunt

* Visible in the browser
visible changes

## Checklist


.storybook/main.ts
```diff
@@ -1,16 +1,15 @@
 import type { AddonOptionsVite } from '@storybook/addon-coverage';
-// @ts-ignore Suppressing type error until types are fixed in react-vite package.
```
andrewgos, line 2: was having 
```
  2:1  error  Use "@ts-expect-error" instead of "@ts-ignore", as "@ts-ignore" will do nothing if the following line is error-free  @typescript-eslint/ban-ts-comment
```

while `@ts-expect-error` was deemed as unused

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Thanks for working on these updates, As mentioned in DMs but added here for clarity:

1. With the single entry point, Does that mean I can only consume tokens along with Bootstrap definitions & component code?
1.1. From my brief test and looking at the build output, this seems to be the case. This would add more content going over the network than required to purely consume the tokens. It could be argued that future component token definitions may reference Bootstrap values in which case we'd need the values here but I am unsure just yet. 
1.2. This approach also diverges from the integration proposal document I generated that mentioned we'd have parallel exports so the end consumer can pick and chose what to import and use.
2. I think that eventually some documentation updates will be required on how the button component can be imported and how the plain tokens can be used in a non React environment.
2.1. Currently the documentation is no longer accurate with this patch applied when doing an import of the locally linked NPM package.
3. Have you tested the build inside of a react app? along with testing importing it into Moodle?
4. I see that there is a failing test relating to the fuzz checking on buttons after your re-arrangement of component files, please address that.

andrewgos:
Thanks for the review @Chocolate-lightning 👍 

What's new:
* SCSS is now generated separately and loaded into the npm build
* Style Dictionary is now made to use the built-in formatter instead of the custom formatter
* Fixed the error with fuzzComponent tests
* Manually tested by running `npm pack` command -> installing into LMS -> copying the build into `public/theme/boost/scss` -> modifying `_default.scss` -> running `grunt` -> resulted in a change in `moodle.css` and was visible in the browser. I will provide screenshots for this in the PR description


I would think that the tidy-up / rearrangement of the CSS loading for the components is implicitly related to getting the design system ready for the "token release", so I'm still leaving it in this PR. But I'm also happy to split it up into a separate PR for another time. Let me know what you think :)

scripts/tokens.ts
```diff
@@ -152,11 +143,20 @@ function convertJsonFileNameToCssFileName(jsonFileName: string) {
   return jsonFileName.replace(/^.*_(.+?)_.*\.json$/, '$1.css').toLowerCase();
 }
 
+/**
```
Chocolate-lightning: Instead of adding a near duplicate function, could we perhaps do something like the following?

```
/**
 * Helper function to convert JSON file name to a target file name (CSS/SCSS).
 * @param jsonFileName
 * @param prefix - e.g. '' for CSS, '_' for SCSS
 * @param extension - e.g. 'css' or 'scss'
 * @returns
 */
function convertJsonFileName(
  jsonFileName: string,
  prefix: string,
  extension: string,
) {
  return jsonFileName
    .replace(/^.*_(.+?)_.*\.json$/, `${prefix}$1.${extension}`)
    .toLowerCase();
}
```

This would ensure that if we can the .replace() filter string that we would not inadvertently miss anything.

scripts/tokens.ts
```diff
@@ -152,11 +143,20 @@ function convertJsonFileNameToCssFileName(jsonFileName: string) {
...
+}
+
 /**
  * Helper function to get file header comment.
  * @param jsonFileName
  * @returns
  */
 function getFileHeader(): string {
-  return `/** THIS FILE IS AUTO-GENERATED — DO NOT EDIT DIRECTLY. **/\n\n`;
+  return `/* Do not edit directly, this file was auto-generated. */\n\n`;
```
Chocolate-lightning: Any particular reason for this change?

tokens/scss/_borders.scss
```diff
@@ -0,0 +1,15 @@
+
```
Chocolate-lightning, line 1: Happens across all .scss files but is there a particular reason why each file includes an initial empty line?

package.json
Chocolate-lightning, line 1: With the changes both here and within the Button component, our [README.md](https://github.com/moodlehq/design-system/blob/main/README.md?plain=1#L111) should be updated to reflect how React components can be pulled in.

tokens/css/index.css
```diff
@@ -1,4 +1,4 @@
-/** THIS FILE IS AUTO-GENERATED — DO NOT EDIT DIRECTLY. **/
+/* Do not edit directly, this file was auto-generated. */
```
Chocolate-lightning: Minor nitpick, This is the only case of an inline comment whereas the other generated css files have it formatted as multi-lined comments.

tokens/scss/_index.scss
```diff
@@ -0,0 +1,10 @@
+/* Do not edit directly, this file was auto-generated. */
```
Chocolate-lightning: Similar to my other nitpicks on CSS & SCSS files but this is the only instance of a SCSS file that does not have a leading empty line.

tokens/scss/_borders.scss
```diff
@@ -0,0 +1,15 @@
...
+// Do not edit directly, this file was auto-generated.
+
+$mds-border-radius-lg: 0.5rem;
```
Chocolate-lightning, line 4: We should perhaps update our README.md to note that whilst CSS token files have reference values for some tokens, the SCSS equivalent files & tokens make use of specified values due to variable hoisting not being available leading to runtime issues.

vite.config.ts
```diff
@@ -18,20 +18,21 @@ export default defineConfig({
...
     rollupOptions: {
-      external: [...Object.keys(peerDependencies)], // Defines external dependencies for Rollup bundling.
+      external: [...Object.keys(dependencies)], // Defines external dependencies for Rollup bundling.
       output: {
         assetFileNames: 'index.css', // name the CSS file
       },
     },
     sourcemap: true, // Generates source maps for debugging.
     emptyOutDir: true, // Clears the output directory before building.
+    minify: false, // Disable minification for JS/CSS/assets
```
Chocolate-lightning, line 35: Do we have some examples or documentation that specifies we should not minify our assets when building the package?

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Thanks for the great work on getting a fully fledged token integration working and tested with the Moodle Core LMS.

I have spotted a few areas that I'd like clarification on, a couple of suggestions to simplify the changes in the diff and a couple of instances of requesting documentation updates.

Please also note the Chromatic checks as there has been a change there with sizing. I don't know if it is a Chromatic thing or if some token values are different / unused now or something else.

Keep it up!

scripts/tokens.ts
andrewgos: very trivial reason, it's to follow the text that came from Style Dictionary's built-in formatter

scripts/tokens.ts
andrewgos: great suggestion! thanks ❤️

tokens/scss/_borders.scss
andrewgos: good question... its coming naturally from SD's code

here they have line separators in the heading
https://github.com/style-dictionary/style-dictionary/blob/main/lib/common/formatHelpers/fileHeader.js#L82-L92

why it didnt happen for css? because [cssVariables](https://github.com/style-dictionary/style-dictionary/blob/main/lib/common/formats.js#L198C1-L201C8) is having `commentStyle.long` while [scssVariables](https://github.com/style-dictionary/style-dictionary/blob/main/lib/common/formats.js#L334) is having `commentStyle.short`

tokens/css/index.css
andrewgos: yes this is because it's generated by our own function getFileHeader().

as i mentioned in the other comment, SD's is having CSS header comment on multi-line. while SCSS is on single line

andrewgos: for consistent sake + for clarity sake (mentioning its generated by style dictionary)

i'll create a custom file header to be applied to all

vite.config.ts
andrewgos: It was a mix of logical arguments from AI and seeing how other design system does it

Logical reasoning: design system is a library -> library should be readable and debuggable -> minification kills that ability to troubleshoot/customize

Though it'll be a different case if we present it as an executable app or static assets

Some examples of other design system npm code
- Google's tokens & components https://www.npmjs.com/package/@material/web?activeTab=code
- IBM's token https://www.npmjs.com/package/@carbon/styles?activeTab=code
- IBM's components https://www.npmjs.com/package/@carbon/web-components?activeTab=code

tokens/scss/_borders.scss
andrewgos: I was contemplating on where should we make a note about those, but good call, README might be a good start 👍 i'll update it

tokens/css/index.css
andrewgos: apparently SD's configuration can only have us customise the text inside, the comment wrapping is beyond our control unless we take over / customize it on the formatter level 😅 so the comment formats needs to stay

package.json
andrewgos: I can confirm that the loading of the components remained the same `import { Button } from '@moodlehq/design-system';`

Thanks to this, I did find an issue with the `.d.ts` gone missing, I've fixed it 👍 

I've also modified the exports in package.json to expose the entrypoints properly, list of entrypoints has been added to the PR descriptions.

tokens/scss/_borders.scss
andrewgos: readme updated :)

package.json
andrewgos: readme updated to display the entrypoints too

andrewgos:
Thanks again for the thorough review @Chocolate-lightning !!

I've addressed each item in the respective inline comments, ready for another review :)

Chocolate-lightning (approved):
Suggestions have been actioned, docs are good, tested it locally with a npm pack then install onto a lightweight React app I have pull in the button component with styles and also tokens inclusion within a CSS file that takes the files we serve in the build.

Looks good, Great work!

PR #145, opened 2026-03-10
feat(Button): Update example component no longer need React-BS

Chocolate-lightning:
## Description

This PR removes the `react-bootstrap` dependency from the `Button` example component and replaces it with a native `<button>` implementation while preserving existing behavior and Bootstrap-compatible class output.

Key changes:
- Replaced `RBButton` usage with native `<button>` in `components/button/Button.tsx`.
- Switched props base type to `ButtonHTMLAttributes<HTMLButtonElement>`.
- Added explicit variant allowlist and fallback to `primary` for invalid variants.
- Kept size support via `btn-sm` / `btn-lg` classes.
- Preserved class composition with `mds-btn`, `btn`, and variant class.
- Removed `ThemeProvider` decorator from `Button` story (no longer needed).
- Removed `react-bootstrap` from `peerDependencies` and `devDependencies`.
- Removed `react-bootstrap` from Vite externals.

https://moodle.atlassian.net/browse/MDS-443

## Type of Change

- [ ] Feature
- [ ] Bugfix
- [x] Refactor
- [ ] Documentation
- [x] Other (dependency removal / package simplification)

## Screenshots/Previews

No visual changes expected beyond removal of `react-bootstrap` runtime dependency.
Storybook `Button` story remains in place and was updated to remove `ThemeProvider`.

## Checklist

## Additional Notes

- Unit tests pass locally with `npm run test-unit` (9/9 passing in `components/button/Button.test.tsx`).
- This change reduces coupling by making the component independent of `react-bootstrap` while keeping bootstrap class conventions.
- `package-lock.json` includes the corresponding dependency graph cleanup.

### DIFF
--- a/components/button/Button.tsx
+++ b/components/button/Button.tsx
@@ -1,16 +1,20 @@
-import {
-  Button as RBButton,
-  ButtonProps as RBButtonProps,
-} from 'react-bootstrap';
+import type { ButtonHTMLAttributes } from 'react';
 
-export interface ButtonProps extends RBButtonProps {
+type ButtonVariant =
+  | 'primary'
+  | 'secondary'
+  | 'danger'
+  | 'outline-primary'
+  | 'outline-secondary'
+  | 'outline-danger';
+
+export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   label: string;
   variant?: string;
-  disabled?: boolean;
   size?: 'sm' | 'lg';
 }
 
-const allowedVariants = [
+const allowedVariants: ButtonVariant[] = [
   'primary',
   'secondary',
   'danger',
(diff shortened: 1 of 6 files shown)


components/button/Button.tsx
```diff
@@ -19,14 +23,30 @@ const allowedVariants = [
...
-export const Button: React.FC<ButtonProps> = ({ label, variant, ...props }) => {
-  const resolvedVariant = allowedVariants.includes(variant ?? '')
-    ? variant
-    : 'primary';
+export const Button = ({
+  label,
+  variant,
+  size,
+  className,
+  type = 'button',
```
andrewgos, line 31: i dont think this attribute is necessary? its always going to be type="button" anyway?

andrewgos (changes requested):
just one minor thing on the prop list

components/button/Button.tsx
Chocolate-lightning: Thanks for the feedback.

I'd rather keep `type` exposed on `ButtonProps`, with a default of `"button"`.

Rationale:
- Native compatibility: exposing `type` preserves expected HTML button behavior.
- Valid form use cases: consumers can explicitly pass `"submit"` or `"reset"` where form actions are needed.

So the component keeps a safe default while still supporting legitimate native button semantics.

andrewgos: ah good call with the other possible types there. in that case, the storybook might need some tidy up there on the `argTypes` section
Screenshot 2026-03-10 at 12 53 51 pm
at the moment its displaying an object instead of string, and description could be added with possible values

andrewgos (approved):
Thanks for tidying up the storybook content! LGTM

PR #177, opened 2026-04-08
feat(components)!: Component library foundations — build, Button API, and authoring guidelines

Chocolate-lightning:
## Description

This PR makes three related improvements:

**1. Component subpath imports**

Each component can now be imported individually via a stable subpath, in addition to the existing full-package import:

```js
// Full package (all components)
import { Button } from '@moodlehq/design-system';

// Component subpath (only Button is fetched)
import { Button } from '@moodlehq/design-system/components/button';

// Direct URL — no build step needed (e.g. Moodle plugins)
import { Button } from 'https://cdn.example.com/mds/dist/components/button/index.js';
```

The build now uses Rollup's `preserveModules` mode — every source file is transpiled in isolation into its own stable, hash-free output file. This matches Moodle's loading model: plugins load files directly by URL without a consumer-side build step, so output paths must remain predictable across releases. Adding a new component folder under `components/` is sufficient to include it in the build — no changes to `package.json` or `vite.config.ts` are needed.

**2. Button variant type narrowing (breaking)**

`ButtonProps.variant` is now typed as `ButtonVariant` (a string union of the six supported values) instead of `string`. Consumers passing an arbitrary string will get a TypeScript error at compile time. Runtime behaviour is unchanged — invalid values still fall back to `"primary"` with a `console.warn` in development.

**3. CI/CD — Chromatic TurboSnap fix**

Corrects Chromatic settings in the testing workflow to ensure TurboSnap functions correctly.

## Type of Change

- [x] Feature
- [x] Bugfix
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist

- [x] Breaking change assessment complete — `ButtonProps.variant` narrowed from `string` to `ButtonVariant`; runtime behaviour unchanged, type errors at compile time for consumers passing arbitrary strings

### DIFF
--- a/.github/instructions/components.instructions.md
+++ b/.github/instructions/components.instructions.md
@@ -24,7 +24,48 @@ export { Button } from './button';
 export type { ButtonProps } from './button';
 ```
 
-Consumers import from the package root only — subpath imports per component are not supported.
+Consumers import from the package root using named imports:
+
+```ts
+import { Button } from '@moodlehq/design-system';
+```
+
+Named re-exports (`export { Button }`) are sufficient for ESM tree-shaking — avoid `export { default as Button }` in barrel files.
+
+## Composition pattern
+
+**Simple components** (single root element, no internal structure): use flat props. Accept all content via named string props rather than `children`. This keeps the API explicit, ensures all text is explicitly caller-supplied, and is straightforward to document.
+
+Two separate rules apply:
+
+1. **Named props over `children`** — simple components do not render `children`; passing children will be silently ignored.
+2. **No raw string literals** — named prop values must be caller-supplied (e.g. from `t()`) so the consuming app controls translation. Raw string literals hardcode English text in the component and bypass the i18n contract.
+
+```tsx
+// ✅ correct — named prop, caller supplies the translated string
+<Button label={t('core:save')} variant="primary" />
+
+// ❌ wrong — raw string literal as a named prop value, not translatable
+<Button label="Save" variant="primary" />
+
+// ❌ wrong — children are not rendered by simple components
+<Button>Save</Button>
+<Button>{t('core:save')}</Button>
+```
+
+**Compound components** (components with distinct named regions — e.g. a card with a header, body, and footer): export each sub-component as a named export from the component's barrel. Do not use dot-notation (`Card.Header`) — named exports are tree-shakeable and align with the existing barrel pattern.
+
+```tsx
+// components/card/index.tsx
+export { Card } from './Card';
+export { CardHeader } from './CardHeader';
+export { CardBody } from './CardBody';
+
+// Consumer
+import { Card, CardHeader, CardBody } from '@moodlehq/design-system';
+```
+
+Each sub-component follows the same file structure rules as any other component and must have its own stories and tests.
 
 ## Breaking change guardrail
 
@@ -51,7 +92,7 @@ import type { ButtonHTMLAttributes } from 'react';
 
 export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   label: string;
-  variant?: string; // intentionally broad — see runtime validation below
+  variant?: ButtonVariant; // use the union type; runtime validation still guards JS consumers
   size?: 'sm' | 'lg';
 }
 ```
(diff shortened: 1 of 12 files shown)


Chocolate-lightning:
## Subpath exports not resolving (`@moodlehq/design-system/button`)

`preserveModules` alone doesn't preserve facade modules (files that only re-export from another module, with a CSS side-effect import that Vite strips from the JS graph). Rollup was optimising them away, so `dist/components/button/index.js` was never generated despite the exports map expecting it.

Fixed by explicitly adding each `components/*/index.tsx` as a Rollup entry point via `rollupOptions.input`, discovered dynamically with `globSync` so new components are picked up automatically. `components/index.tsx` was also changed to import directly from `./button/Button` rather than `./button` to avoid Rollup creating a duplicate `Button2.js` when the same module was reachable from two entry chains.

## TypeScript error on `import '@moodlehq/design-system/css'`

The `./css` export had no `types` condition. With `moduleResolution: "bundler"` and `noUncheckedSideEffectImports: true` in consuming projects, this is a hard error. Added `css.d.ts` at the repo root and wired it into the exports map and `files`.

## Chromatic showing unstyled components

`sideEffects: ["dist/index.css"]` marked `index.ts` as side-effect-free. When Rollup builds the Storybook static site for Chromatic, it was dropping `import '../index'` in `preview.ts` (no bindings used, module appears side-effect-free) — taking Bootstrap, the design tokens, and component CSS with it.

Fixed on two fronts: `sideEffects` updated to `["**/*.css"]` so CSS files are never tree-shaken, and `preview.ts` updated to import Bootstrap, tokens, and component CSS directly rather than relying on `index.ts` as a side-effect carrier.

.github/instructions/components.instructions.md
```diff
@@ -24,7 +24,40 @@ export { Button } from './button';
...
+## Composition pattern
+
+**Simple components** (single root element, no internal structure): use flat props. Accept all content via props — no `children`. This keeps the API explicit, i18n-safe (all strings are caller-supplied), and straightforward to document.
+
+```tsx
+// ✅ correct for simple components
+<Button label={t('core:save')} variant="primary" />
+
+// ❌ wrong — children bypass the i18n contract
+<Button>Save</Button>
```
andrewgos: im not fully understanding this one, isnt also possible to do i18n through the children?

`<Button>{t('core:save')}</Button>`

css.d.ts
```diff
@@ -0,0 +1,3 @@
+// Type declaration for the CSS side-effect import.
+// import '@moodlehq/design-system/css'
```
andrewgos: do we need this?

package.json
```diff
@@ -5,14 +5,20 @@
...
-    "/tokens/scss"
+    "/tokens/scss",
+    "/css.d.ts"
   ],
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
     },
+    "./*": {
```
andrewgos: After testing the output and checking further, AI can testify that subpath imports aren't necessary

With the help of ESM named imports, tree-shaking are done as long as it is a named export and not a default export.

```
// ✅ Good for tree-shaking
export { Button } from './button';

// ❌ Bad for tree-shaking
export { default as Button } from './button';
```

This can be something to improve the instructions of the AI

.github/instructions/components.instructions.md
Chocolate-lightning: Good catch — the wording is too broad. `<Button>{t('core:save')}</Button>` is i18n-safe because `t()` is a caller-supplied translated string, not a hardcoded literal.

The actual concern is:

`<Button>Save</Button>` — raw string literal, not translatable, bypasses the i18n contract ❌
`<Button>{t('core:save')}</Button>` — translated string passed as children, still caller-supplied ✅
The instruction has been updated to reflect this distinction: children are not banned outright; raw string literals as children are the anti-pattern.

css.d.ts
Chocolate-lightning: Yes — this file is required for TypeScript consumers using `moduleResolution: "bundler"` with `noUncheckedSideEffectImports: true`. 
Without a types condition on the `./css` export, `import '@moodlehq/design-system/css'` is a hard compile-time error in those projects, I tested this with my small testing React application.

The comment in the file has been improved to explain this context directly.

Chocolate-lightning: You're right. Named re-exports (`export { Button }`) combined with ESM are sufficient for bundlers (Vite, webpack 5, Rollup, esbuild) to tree-shake unused components — subpath imports add no benefit here. The key rule is to avoid `export { default as Button }`, which the barrel already doesn't use.

Since the `"./*"` wildcard export was added in this PR and has no consumers, it has been removed along with the coupled Vite build changes:

- `package.json` — removed the `"./*"` wildcard exports entry
- `vite.config.ts` — removed `rollupOptions.input` glob, `preserveModules`, `preserveModulesRoot`, `entryFileNames`; restored `fileName: () => 'index.js'` in the lib config
- `components/index.tsx` — reverted imports back to `'./button'` (the direct-to-implementation path was only needed to avoid Rollup duplicate-module warnings when component barrels were explicit entry points)

package.json
andrewgos: Resolved as answered in the other comment thread

.github/instructions/components.instructions.md
andrewgos: thanks for clarifying! though i believe we do not support injecting through children yet? so the example may not be applicable there.

i'd imagine it'll be more like

```
// ✅ correct — named prop, caller supplies the translated string
<Button label={t('core:save')} variant="primary" />

// ❌ wrong — raw string literal, not translatable
<Button label="Save" variant="primary" />

// ❌ wrong — raw string literal as children not supported
<Button>Save</Button>
```

depends on what message that this section wanted to convey. Is it more about the translation, or about the named prop?

Chocolate-lightning: Good call out!

I updated the instruction files to better reflect what is desired.

andrewgos (approved):
Thanks for getting the subpath import back in! The updated instructions also look good.

LGTM ✅

PR #178, opened 2026-04-10
feat(icons): Add fontawesome for storybook

andrewgos:
## Description

- Added `@fortawesome/fontawesome-free` package as a dev dependency
- It is not bundled into the published package - with the purpose of having the consumer's freedom to choose their own icon providers.
- Button components are updated swiftly
  - To show the intended usage of the Fontawesome package in the repository
  - Considered to have as props to be strict on the icon positioning, while opening the option for any icons injected
  - CSS changes are minimal and will be revisited in https://moodle.atlassian.net/browse/MDS-266
- Unit test updated slightly to accommodate the situation where the label of the button would be dynamic 

https://moodle.atlassian.net/browse/MDS-480

## Type of Change

- [x] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Screenshots/Previews
Screenshot 2026-04-10 at 11 30 30 am

## Checklist

## Additional Notes
N/A

### DIFF
--- a/components/button/Button.stories.tsx
+++ b/components/button/Button.stories.tsx
@@ -2,6 +2,16 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
 import { expect } from 'storybook/test';
 import { Button } from './Button';
 
+const iconMapping = {
+  None: undefined,
+  Download: <i className="fa-solid fa-download" aria-hidden="true" />,
+  Trash: <i className="fa-solid fa-trash" aria-hidden="true" />,
+  'Arrow Right': <i className="fa-solid fa-arrow-right" aria-hidden="true" />,
+  'Arrow Left': <i className="fa-solid fa-arrow-left" aria-hidden="true" />,
+  Plus: <i className="fa-solid fa-plus" aria-hidden="true" />,
+  Check: <i className="fa-solid fa-check" aria-hidden="true" />,
+};
+
 const meta = {
   title: 'Components/Button',
   component: Button,
@@ -9,10 +19,10 @@ const meta = {
     layout: 'centered',
   },
   play: async ({ canvas, userEvent }) => {
-    await userEvent.click(canvas.getByRole('button', { label: 'Button' }));
+    await userEvent.click(canvas.getByRole('button'));
     // Wait for any updates to complete
     await new Promise((resolve) => setTimeout(resolve, 0));
-    await expect(canvas.getByText('Button')).toBeVisible();
+    await expect(canvas.getByRole('button')).toBeVisible();
   },
   tags: ['autodocs', 'test', 'beta'],
   // More on argTypes: https://storybook.js.org/docs/api/argtypes
@@ -50,6 +60,20 @@ const meta = {
         type: { summary: 'sm | lg' },
         defaultValue: { summary: 'undefined' },
       },
+    },
+    startIcon: {
+      description:
+        'Icon rendered before the label. Accepts only intrinsic `<i>` or `<svg>` elements. Mark decorative icons with `aria-hidden="true"`. For icon-only buttons pass `aria-label` directly.',
+      options: Object.keys(iconMapping),
+      mapping: iconMapping,
+      control: { type: 'select' },
+    },
+    endIcon: {
+      description:
+        'Icon rendered after the label. Accepts only intrinsic `<i>` or `<svg>` elements. Mark decorative icons with `aria-hidden="true"`. For icon-only buttons pass `aria-label` directly.',
+      options: Object.keys(iconMapping),
+      mapping: iconMapping,
+      control: { type: 'select' },
     },
     disabled: {
       control: { type: 'boolean' },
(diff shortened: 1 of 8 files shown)


components/button/Button.tsx
```diff
@@ -46,7 +50,9 @@ export const Button = ({
 
   return (
     <button className={classes.join(' ')} type={type} {...props}>
+      {startIcon}
```
Chocolate-lightning: No DEV warning when label is absent and no aria-label is provided. Silent inaccessible button. Worth a console.warn in import.meta.env.DEV.

Chocolate-lightning: Also — should startIcon and endIcon be mutually exclusive? Nothing in ZeroHeight or the instruction files says. If yes, the type or a DEV warning should enforce it. If no, the combined story and tests above are needed. We may need to chat about this in the prod-ready button issue.

components/button/Button.test.tsx
```diff
@@ -40,6 +40,32 @@ describe('Button: Unit Test', () => {
...
+        label="Button"
+        startIcon={<span data-testid="start-icon">S</span>}
+        endIcon={<span data-testid="end-icon">E</span>}
+      />,
+    );
+
+    const button = screen.getByRole('button');
+    expect(screen.getByTestId('start-icon')).toBeInTheDocument();
+    expect(screen.getByTestId('end-icon')).toBeInTheDocument();
+    expect(button).toHaveTextContent('SButtonE');
```
Chocolate-lightning: This checks concatenated text content, not DOM order. Whilst it fails correctly if you manually swap the start & end icons in the button, an assert order via childNodes index comparison should be done instead.

components/button/Button.test.tsx
```diff
@@ -40,6 +40,32 @@ describe('Button: Unit Test', () => {
...
+  });
+
+  it('supports icon-only usage with an aria-label', () => {
```
Chocolate-lightning, line 66: Test only checks the button is findable by name. Doesn't assert that no label text appears in the DOM — worth adding given label defaults to 'Button' in meta.

components/button/Button.test.tsx
```diff
@@ -40,6 +40,32 @@ describe('Button: Unit Test', () => {
...
+    render(
+      <Button
+        startIcon={<span data-testid="start-icon">S</span>}
+        aria-label="Delete"
+      />,
+    );
+
+    expect(screen.getByRole('button', { name: 'Delete' })).toBeInTheDocument();
+  });
```
Chocolate-lightning, line 171: A few cases may be worth adding:

```
startIcon alone (no endIcon)
endIcon alone (no startIcon)
Neither prop — no unexpected markup in the button
label + startIcon present — icon content doesn't pollute the accessible name
```

components/button/Button.stories.tsx
```diff
@@ -51,6 +51,16 @@ const meta = {
         defaultValue: { summary: 'undefined' },
       },
     },
+    startIcon: {
```
Chocolate-lightning, line 64: Descriptions say <i> or SVG but both props accept any ReactNode. Also worth making the aria-hidden="true" requirement more prominent — easy to miss as prose.

components/button/Button.stories.tsx
```diff
@@ -149,3 +159,24 @@ export const Small = {
     label: 'Button',
   },
 } satisfies Story;
+
```
Chocolate-lightning, line 176: Our instructions & checklist require an RTL story for any component with directional icon placement. Need a story wrapping the icon variants in dir="rtl" to verify CSS logical properties mirror correctly.

Chocolate-lightning (changes requested):
Kia ora Andrew,

Overall, nice work here and is inline with what I have been experimenting with locally but a bit less opinionated which is better for end consumers.

Just a few items to address mainly around testing but one major one for i18n.

Cheers,

components/button/Button.stories.tsx
```diff
@@ -149,3 +159,24 @@ export const Small = {
...
+} satisfies Story;
+
+export const WithTrailingIcon: Story = {
+  args: {
+    label: 'Continue',
+    endIcon: <i className="fa-solid fa-arrow-right" aria-hidden="true" />,
+  },
+} satisfies Story;
+
+export const IconOnly: Story = {
```
Chocolate-lightning, line 191: One more point sorry, When running storybook locally, The text appears. You can also see it happening here: https://www.chromatic.com/review?appId=6901bebddeae4c83b504fdad&number=178&type=linked&activeElementId=comparison-test-key-69d853bea8d7155814436c20-1200px&view=changes

Chocolate-lightning (changes requested):

components/button/Button.stories.tsx
andrewgos: made it to be  `type IconElement = ReactElement<'i' | 'svg'>;` to enforce better.

any suggestions on what more prominent may look like?

components/button/Button.stories.tsx
andrewgos: added one in

components/button/Button.stories.tsx
andrewgos: fixed, it was defaulting to the meta-level default args of storybook

components/button/Button.test.tsx
andrewgos: good catch! updating the asserts

components/button/Button.test.tsx
andrewgos: added assertion to double check the textcontent

components/button/Button.test.tsx
andrewgos: good idea. added those in

components/button/Button.tsx
andrewgos: added check to ensure a label access is available

with the mutual exclusivity, there isnt any requirements that came from the designers, so i think lets keep it open for the user to set it up, since both startIcon and endIcon are not interfering with one another (except that it may looks weird..)

Chocolate-lightning (approved):
Everything looks good for the start of button development. Rebased to the latest as well, so merging now.

PR #179, opened 2026-04-10
feat(Tokens): Add new activity icon tokens and modify sizing values

borisbasque:


Chocolate-lightning (approved):
Tokens updated as found in ZH and I have also added a small update to the button to reference the token that has the same value as it had before.

andrewgos (approved):
LGTM 🚢

PR #185, opened 2026-04-20
feat(Radio): Include new Radio input component

Chocolate-lightning:
## Description

Adds the `Radio` component to the design system.

The component is a single-input form control that wraps a native `<input type="radio">` with a visible label, optional invalid feedback text, and a `hideLabel` mode for label-less contexts. It is implemented using Bootstrap CSS class hooks and `--mds-*` tokens, following the same pattern as `Button`.

Key implementation decisions:

- **`invalid` and `invalidFeedback` are independent props** — `invalid` controls styling and `aria-invalid`; `invalidFeedback` controls the visible feedback message. `invalidFeedback` requires `invalid={true}` to render; a dev warning fires if `invalidFeedback` is supplied without `invalid`.
- **`hideLabel` suppresses `invalidFeedback`** — feedback text requires a visible label to provide context. When `hideLabel={true}`, `invalidFeedback` is ignored and a dev-mode warning is emitted.
- **`aria-invalid` is set explicitly** on the input when `invalid={true}`, rather than relying on the Bootstrap `is-invalid` class alone, which screen readers cannot detect.
- **`forwardRef`** — the component forwards a ref to the underlying `<input>` element, enabling form library integration and programmatic focus management. `Radio.displayName` is set.
- **`mds-*` hook classes on all elements** — `mds-form-check` (wrapper), `mds-form-check-input` (input), `mds-form-check-label` (label), `mds-form-check-feedback` (feedback div). Bootstrap classes are also applied for base styling.
- **Grid layout pins radio to top of multi-line labels** — `align-items: start` on the wrapper matches the Figma spec; for single-line labels this is visually identical to centred alignment.
- **Focus ring uses a two-layer `box-shadow`** — a 1px white inner ring (`--mds-stroke-weight-sm`) creates a gap between the circle and the 2px coloured focus indicator. This is an intentional deviation from Figma (which has no gap) for improved legibility against varied backgrounds.
- **No `is-valid` or neutral feedback states** — only `invalid` is supported. Radio buttons surface a binary result; positive confirmation is conveyed by form-level success messaging rather than per-input valid styling.
- **No hover styles** — the radio input relies on OS/browser default appearance for hover. No design token exists for a radio hover state and none has been requested.
- **Disabled state opacity on feedback** — `opacity: 0.5` is applied via `:has(.mds-form-check-input:disabled)` on the feedback element. No `--mds-opacity-*` token currently exists; this value should be updated to a token once one is defined.

---

https://moodle.atlassian.net/browse/MDS-498

---

## Screenshots/Previews

| File | Shows |
| ---- | ----- |
| feedback-invalid | Invalid state — danger border, label in danger colour, feedback text below |
| disabled-checked | Disabled + checked — muted blue fill, muted label; confirms disabled token values |
| group | Horizontal and vertical group layouts — mutual exclusion via shared `name` |
| right-to-left | RTL layout — Arabic label and feedback, radio appears on the trailing (right) edge |
| multiline-label | Multi-line label — radio pinned to top of text (`align-items: start`); matches Figma spec |
| focus-ring-checked | Focus ring on checked state — two-layer box-shadow, white gap visible between circle and 2px indicator |
| focus-ring-invalid | Focus ring on invalid state — danger colour ring, confirms 2px width (not Figma's erroneous 3px) |

---

- [x] All 5 required files present: `Radio.tsx`, `radio.css`, `Radio.test.tsx`, `Radio.stories.tsx`, `index.tsx`

### TypeScript and props

- [x] Props interface extends the correct React HTML element interface (`InputHTMLAttributes<HTMLInputElement>`)
- [x] Constrained props use `allowedValues` runtime validation with a graceful fallback to a documented default — N/A: `Radio` has no constrained multi-value props (no `variant` or `size`). DEV warnings are used instead for props that have conditional rendering requirements (`invalidFeedback` without `invalid`, `hideLabel` without an accessible label).
- [x] `...props` spread last on the root element

### Internationalisation

### CSS and tokens

- [x] All CSS values use `var(--mds-*)` tokens — no hardcoded colours, spacing, or typography. **Exception:** `opacity: 0.5` on disabled feedback — no `--mds-opacity-*` token exists yet; tracked in description above.

### Tests and stories

- [x] Storybook stories cover all documented variants, sizes, and states (11 stories; all include JSDoc)

### Breaking changes

### AI context

- [x] Instruction files updated — four patterns added to `components.instructions.md` and `stories-tests.instructions.md`: `forwardRef` requirement for focusable-element components (including `displayName`); `mds-*` hook on every styled inner element (not just root); `aria-describedby` linkage pattern for input/feedback pairs; ref forwarding added to unit test baseline checklist

---

## Cross-system parity

---

### Branch history

> This branch is currently based off of another option PR that aims to improve various facetes of the underlaying component development pipeline, thus this PR should not be merged ahead of https://github.com/moodlehq/design-system/pull/177

### Accessibility

- `aria-invalid="true"` is set on the input directly when `invalid={true}`. Screen readers use this attribute, not the CSS class.
- `aria-describedby` links the input to its feedback element when both `invalid` and `invalidFeedback` are set, so screen readers announce the error message.
- `aria-label` is derived from the `aria-label` prop → `label` prop fallback when `hideLabel={true}`. A dev warning fires if neither is available.
- A dev warning fires when `label` is absent and `hideLabel` is false — an empty `<label>` is an accessibility issue.
- All stories pass axe WCAG 2.x A/AA automated checks.

### Storybook a11y — `region` rule disabled globally (`.storybook/preview.ts`)

The axe `region` best-practice rule has been disabled in `options.rules`:

```ts
options: {
  rules: { region: { enabled: false } },
}
```

This rule requires all page content to sit inside a landmark (`<main>`, `<header>`, etc.). Component stories are structurally never pages, so the rule produces a false positive on every story in the design system. This is a documented pattern in the Storybook a11y docs.

`options.rules` is used rather than `config.rules` because `config` maps to `axe.configure()` (global, not reset per story), while `options` maps to `axe.run()` (fresh on every render).

**Coverage impact:** None. The rule is `best-practice` only. All WCAG 2.x A/AA checks remain fully enabled.

### Stories added

All 11 stories include JSDoc documentation describing their purpose and usage context.

| Story                 | Purpose                                                            |
| --------------------- | ------------------------------------------------------------------ |
| `Default`             | Click-to-select interaction                                        |
| `Checked`             | Pre-checked state                                                  |
| `Group`               | Horizontal and vertical radio group layouts                        |
| `Disabled`            | Disabled state                                                     |
| `DisabledChecked`     | Disabled + checked combination                                     |
| `InvalidNoFeedback`   | Invalid styling without feedback text                              |
| `FeedbackInvalid`     | Invalid styling with feedback message                              |
| `InputOnly`           | `hideLabel` with explicit `aria-label`                             |
| `HideLabelFallback`   | `hideLabel` with `label` prop as `aria-label` fallback             |
| `KeyboardInteraction` | Tab to focus, Space to select                                      |
| `RightToLeft`         | RTL layout verification (Arabic label and feedback; no Figma spec) |

### DIFF
--- a/.github/instructions/components.instructions.md
+++ b/.github/instructions/components.instructions.md
@@ -141,6 +141,8 @@ For components built on a Bootstrap base class (e.g. `btn`): include the Bootstr
 
 For components not built on Bootstrap: the `mds-*` hook is still required; omit Bootstrap classes entirely.
 
+**Apply an `mds-*` hook class to every element that has component-scoped CSS rules**, not just the root. This allows consumers and the CSS to target any element directly without relying on descendant selectors alone. For example, a component with a wrapper, an input, a label, and a feedback element should apply `mds-form-check`, `mds-form-check-input`, `mds-form-check-label`, and `mds-form-check-feedback` respectively. Bootstrap classes are applied alongside the `mds-*` hooks where needed for base styling.
+
 ```tsx
 // Example: component with Bootstrap base
 const classes = ['mds-btn', 'btn', `btn-${resolvedVariant}`];
@@ -155,6 +157,23 @@ return (
 ```
 
 Spread `...props` last so consumers can pass `aria-*`, `data-*`, event handlers, and other native attributes through without explicit forwarding.
+
+## Ref forwarding
+
+Any component that renders a native focusable element (`<input>`, `<button>`, `<textarea>`, `<select>`, `<a>`) must use `forwardRef` so that form libraries (React Hook Form, Formik) and consumers that need programmatic focus management can access the underlying DOM node.
+
+```tsx
+import { forwardRef } from 'react';
+
+export const Radio = forwardRef<HTMLInputElement, RadioProps>(
+  ({ label, ...props }, ref) => (
+    <input ref={ref} type="radio" {...props} />
+  ),
+);
+Radio.displayName = 'Radio'; // required — forwardRef components lose their inferred name
+```
+
+Always set `ComponentName.displayName` explicitly. Without it the component appears as `ForwardRef` in React DevTools and error messages.
 
 ## Internationalisation
 
@@ -185,6 +204,23 @@ Direction-neutral properties (`top`, `bottom`, `height`, `width`, `margin-top`, 
 **`dir` attribute:** No explicit forwarding needed — writing direction is inherited from the document or nearest ancestor. Because `...props` is always spread on the root element, consumers can pass `dir` directly if needed.
 
 **Locale-aware formatting:** This library is presentation-only and does not format dates, numbers, or currency. Components accept pre-formatted strings; locale-aware formatting is the consumer's responsibility.
+
+## Accessibility wiring for feedback/description elements
+
+When a component renders a visible helper or error message alongside an input, link them with `aria-describedby` so screen readers announce the message when the field receives focus.
+
+- Generate a stable ID for the message element derived from the input's ID (e.g. `${id}-feedback`).
+- Set `aria-describedby` on the input pointing to that ID.
+- Only generate the feedback ID and render the element when the message will actually be displayed — do not leave empty elements with IDs in the DOM.
+
+```tsx
+const feedbackId = invalid && invalidFeedback ? `${id}-feedback` : undefined;
+
+<input aria-describedby={feedbackId} ... />
+{feedbackId && (
+  <div id={feedbackId}>{invalidFeedback}</div>
+)}
+```
 
 ## Agent guardrails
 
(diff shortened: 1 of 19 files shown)


.github/instructions/components.instructions.md
```diff
@@ -226,3 +301,43 @@ Component styles layer on top of Bootstrap CSS classes. The JSX applies both the
 Never use `!important` — always resolve conflicts through specificity or cascade order.
 
 Token naming follows `--mds-{category}-{subcategory}-{modifier}` — for example `--mds-bg-interactive-primary-default`, `--mds-spacing-xs`, `--mds-font-size-paragraph-default`. See `tokens/css/` for the full set.
+
+## Focus ring pattern
+
+All MDS components must use `outline` + `outline-offset` for focus rings, **not** `box-shadow`. This is a deliberate, documented divergence from Bootstrap's default focus ring pattern.
```
andrewgos: shall we mentioned **not** `border` too? in case its trying to do funny tricks with data from Figma

components/radio/index.tsx
```diff
@@ -0,0 +1,3 @@
+import './radio.css';
```
andrewgos, line 3: is this for css tree-shaking purposes? 🤔  i dont think ive seen this on the `button` yet when we did the foundational PR

components/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,371 @@
...
+        type: {
+          summary: 'true | false',
+        },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    hideLabel: {
+      control: { type: 'boolean' },
+      description:
+        'When true, hides the visible label element. The input will be labelled by aria-label (if provided) or by the label prop as a fallback. Use this when a visible label is not desired but accessibility must be maintained.',
```
andrewgos: pardon my ignorance, in what cases do we need to hide labels for the radio button? 🤔

components/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,371 @@
...
+      control: { type: 'boolean' },
+      description:
+        'Marks the input as invalid, applying danger styling to the input border and label. Set this independently of invalidFeedback to show invalid state without a feedback message.',
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+  },
+  args: {
```
andrewgos, line 127: probably should add `checked`, `hideLabel`  in this list too, so that it doesnt appear as "Set boolean" in the control

components/radio/Radio.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+    expect(screen.getByRole('radio')).toBeChecked();
+  });
+
+  it('forwards a ref to the underlying input element', () => {
+    const ref = { current: null as HTMLInputElement | null };
+    render(<Radio label="Option" ref={ref} />);
+    expect(ref.current).toBeInstanceOf(HTMLInputElement);
+  });
+
+  it('applies className to the wrapper', () => {
```
andrewgos, line 88: is this description correct? 🤔  i suppose wrapper is the root and not the `<input>`

andrewgos (commented):
overall looks good! just a few minor items and questions :)

.github/instructions/components.instructions.md
Chocolate-lightning: The instructions file now explicitly states that neither box-shadow nor border should be used for focus rings. 

The CSS comments direct contributors to the instructions file for the full rationale. Only outline + outline-offset are permitted for accessibility and forced colors compliance.

components/radio/Radio.stories.tsx
Chocolate-lightning: The JSDoc for the hideLabel prop now explains valid use cases, such as dense tables where the column header acts as the label or icon-only UIs. 

It also emphasizes that an accessible name must always be provided via aria-label or the label prop.

components/radio/Radio.test.tsx
Chocolate-lightning: The test description has been clarified to state "applies className to the root wrapper element" for accuracy.

components/radio/Radio.stories.tsx
Chocolate-lightning: Storybook controls have been updated to explicitly list checked and hideLabel with clear descriptions, improving the controls panel and documentation.

components/radio/index.tsx
Chocolate-lightning: I have added a comment to clarify that the CSS import ensures local styles are included with subpath imports, matching the intended pattern for tree-shaking and consistency across components.

andrewgos (dismissed):
Thanks for answering my questions and improving the docs! LGTM 🚀

andrewgos (approved):
🛳️

PR #191, opened 2026-04-22
feat(close-button): Add new CloseButton component

andrewgos:
## Description

Adds the CloseButton component, a compact and accessible dismiss control built on Bootstrap’s btn-close pattern.

The component renders a semantic button with a required aria-label prop (no default), supports three sizes (`sm`, `md`, `lg`), and defaults to `md` when size is missing or invalid. It accepts all native `ButtonHTMLAttributes<HTMLButtonElement>` props and forwards them to the host element. Styling is implemented with Moodle design tokens only (`--mds-*` custom properties), with no hardcoded color, spacing, typography, or border values.

**Key implementation decisions:**

- Uses Bootstrap `btn-close` for baseline close-icon rendering and behavior consistency with existing Bootstrap-based UI patterns.
- Adds a stable root hook class (`mds-close-button`) plus size modifiers (`mds-close-button--sm`, `mds-close-button--md`, `mds-close-button--lg`) to keep styling deterministic and testable.
- Uses runtime `allowedSizes` validation to guard size and gracefully fall back to `md` instead of throwing or failing silently.
- Requires `aria-label` to enforce i18n compliance; consuming applications must pass translated/contextual text.
- Locks type to button at compile time (`Omit<..., "type">`) and runtime (hard-coded after prop spread) to prevent accidental form submission.
- Spreads `...props` on the button to preserve native extensibility (`disabled`, `onClick`, `data-*`, `aria-*`, etc.) and ensure drop-in compatibility.
- The focus ring is done with the `outline` CSS attribute.
- No `:active` / `:focus` states upon clicking, only `:focus-visible` for accessibility on keyboard selector

https://moodle.atlassian.net/browse/MDS-495

## Screenshots/Previews
Screenshot 2026-04-23 at 9 57 39 am
Screenshot 2026-04-23 at 9 58 04 am

### TypeScript and props

- [x] `...props` spread last on the primary (interactive) element

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes

N/A


components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+   * Visual size variant for the close icon button.
+   * Invalid values fall back to the default `md` size at runtime.
+   */
+  size?: string;
+}
+
+const allowedSizes: CloseButtonSize[] = ['sm', 'md', 'lg'];
+
+export const CloseButton = ({
+  'aria-label': ariaLabel = 'Close',
```
Chocolate-lightning: I don't think we should be adding in a hardcoded language string here. I believe we should warn the developer if they are invoking this component and are missing the label value.

components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,45 @@
+import type { ButtonHTMLAttributes } from 'react';
+
+type CloseButtonSize = 'sm' | 'md' | 'lg';
+
+export interface CloseButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+  /**
+   * Accessible name announced by screen readers for the close button control.
+   */
+  'aria-label'?: string;
```
Chocolate-lightning: This is a redundant definition and paired with the next comment in this file, is a i18n fail. The caller should define this value at invocation.

components/close-button/CloseButton.test.tsx
```diff
@@ -0,0 +1,75 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { CloseButton } from './CloseButton';
+
+describe('CloseButton: Unit Test', () => {
+  it('applies mds, Bootstrap, and default md size classes', () => {
```
Chocolate-lightning: The current size tests (`applies sm size class`, `applies lg size class`) could be simiplifed into a single `it.each` over `['sm', 'md', 'lg']`, with the base-class assertions (`mds-close-button`, `btn-close`) included in the body — this gives you coverage of those classes across all three variants rather than just `md` and then running two seperate size tests. 

The invalid-value fallback test is fine remaining seperate as it is testing guardrails.

components/close-button/CloseButton.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  it('applies lg size class', () => {
+    render(<CloseButton size="lg" />);
+    expect(screen.getByRole('button', { name: 'Close' })).toHaveClass(
+      'mds-close-button--lg',
+    );
+  });
+
+  it('falls back to md size class for an invalid size value', () => {
+    render(
```
Chocolate-lightning: You could likely do something like the following as a simple readability improvement:

```
// eslint-disable-next-line @typescript-eslint/no-explicit-any
render(<CloseButton size={'invalid' as any} />);
```

Rather than having the eslint definition within the passed object.

components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ];
+  if (className) {
+    classes.push(className);
+  }
+
+  return (
+    <button
+      className={classes.join(' ')}
+      aria-label={ariaLabel}
+      {...props}
```
Chocolate-lightning, line 56: The `<button>` does not have an explicit `type` attribute, so it defaults to `"submit"` in the spread. If this component were to be used in a form context , It would trigger the submission of the form if the implementor did not catch it. 

Maybe we could add `type="button"` explicitly on the element and add appropriate testing for the value. Check here for more info: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#type

components/close-button/CloseButton.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { CloseButton } from './CloseButton';
+
+const meta = {
+  title: 'Components/CloseButton',
+  component: CloseButton,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'beta'],
+  argTypes: {
```
Chocolate-lightning, line 12: The component spreads `...props` onto the `<button>`, so native event handlers like `onClick` work — but because autodocs only surfaces explicitly declared `argTypes`, they're invisible in the Storybook props table. Consider adding an `onClick` entry to `argTypes` so consumers can discover it from the docs without reading the source:

```tsx
onClick: {
  description: 'Callback fired when the button is clicked. Receives the native mouse event.',
  table: {
    type: { summary: '(event: React.MouseEvent<HTMLButtonElement>) => void' },
  },
},
```

.github/PULL_REQUEST_TEMPLATE/component.md
```diff
@@ -23,7 +23,7 @@
 - [ ] Props interface extends the correct React HTML element interface (e.g. `ButtonHTMLAttributes<HTMLButtonElement>`)
 - [ ] JSDoc comment on each prop
 - [ ] Constrained props use `allowedValues` runtime validation with a graceful fallback to a documented default
-- [ ] `...props` spread last on the root element
+- [ ] `...props` spread last on the primary (interactive) element
```
Chocolate-lightning: I am a bit unsure about the usage of (interactive) here. Mainly with thoughts of the upcoming activity icon component which by definition is not interactive.

Keen to hear your thoughts on this.

.github/instructions/stories-tests.instructions.md
Chocolate-lightning, line 1: See: https://github.com/moodlehq/design-system/pull/191/changes#r3136369022

.github/instructions/components.instructions.md
Chocolate-lightning, line 1: See: https://github.com/moodlehq/design-system/pull/191/changes#r3136369022

Chocolate-lightning (changes requested):
Kia ora Andrew,

Overall, great work on this component. I have spotted a few items that I think should be addressed as well as some areas for potential improvement.

Take a ganger over the points and if you have any questions, let me know.

Cheers,

components/close-button/CloseButton.tsx
andrewgos: Good point. requiring a prop that is an accessibility-related feels unusual, seeing react-boostrap defaulting it, but it's worth enforcing indeed 👍

components/close-button/CloseButton.tsx
andrewgos: as resolved on the next comment, making it as a required field

components/close-button/CloseButton.tsx
andrewgos: Very good catch!! it is dangerous indeed. enforcing the type "button" after the prop spread and unit tests to ensure no disruptions from the prop spread

components/close-button/CloseButton.test.tsx
andrewgos: Consulted with AI further, perhaps casting `as any` is not adding much value since it is already typed as a string. Cleaning this up

.github/PULL_REQUEST_TEMPLATE/component.md
andrewgos: agree on that 👍  after considering some other synonyms, perhaps we can call it the "host element" ?

components/close-button/CloseButton.test.tsx
andrewgos: taking this in 👍

.github/instructions/components.instructions.md
andrewgos: renamed to "host element"

.github/instructions/stories-tests.instructions.md
andrewgos: renamed to "host element"

.github/instructions/components.instructions.md
```diff
@@ -80,7 +80,7 @@ Prop mapping conventions:
 
 ## Composition pattern
 
-**Simple components** (single root element, no internal structure): use flat props. Accept all content via named string props rather than `children`. This keeps the API explicit, ensures all text is explicitly caller-supplied, and is straightforward to document.
+**Simple components** (single host element, no internal structure): use flat props. Accept all content via named string props rather than `children`. This keeps the API explicit, ensures all text is explicitly caller-supplied, and is straightforward to document.
```
Chocolate-lightning, line 83: I think host element is a fine name, Bounding element would have also sufficed.

components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface CloseButtonProps extends Omit<
+  ButtonHTMLAttributes<HTMLButtonElement>,
+  'type'
+> {
+  /**
+   * Accessible name announced by screen readers for the close button control.
+   * Must be a translated string provided by the caller.
+   */
+  'aria-label': string;
```
Chocolate-lightning, line 13: Given that aria-label is a part of the underlaying AriaAttributes interface in React,  It will naturally be spread out in the ...props. I was thinking that maybe this prop could be omitted from the Close Button definition, but given that there will never be visible text in this component and we explicitly need a provided value from the consumer, we can leave this as is.

Just wanted to comment for prosperity.

components/close-button/close-button.css
Chocolate-lightning, line 1: I noticed that the ```--mds-spacing-offset``` token is never used in your definitions, was it never needed for this component?

Chocolate-lightning (changes requested):
I think once we get CC set up for this and square away a few small remaining points, we could look at landing this before the end of the week and shipping a release with two components included.

components/close-button/close-button.css
andrewgos: yes, just as @borisbasque shared with me, we no longer have a container wrapping the icon, so no offset is needed there for the close button

Chocolate-lightning (approved):
All changes have been actioned, Designs have been signed off by designers.

Accepting this.

PR #198, opened 2026-04-29
ci: Avoid npm binding issues in SB deploy

Chocolate-lightning:
## Description

The CI build was failing with the error:
```
Error: Cannot find native binding. npm has a bug related to optional dependencies.
Please try `npm i` again after removing both package-lock.json and node_modules directory.
```

### Root Cause
The committed `package-lock.json` contains only macOS rolldown bindings (`@rolldown/binding-darwin-x64`), but Linux CI requires `@rolldown/binding-linux-x64-gnu`. When using `npm ci` (strict lockfile mode), npm refuses to install because the required platform binding is not in the lockfile for that platform.

### Solution
Regenerated `package-lock.json` from a clean install so the lockfile includes the required optional native rolldown bindings for both macOS and Linux.

### Validation
- Regenerated lockfile now includes:
	- `@rolldown/binding-darwin-arm64`
	- `@rolldown/binding-darwin-x64`
	- `@rolldown/binding-linux-x64-gnu`
- Local Storybook production build succeeds with strict lockfile flow: `npm run build-storybook`

## Related Jira Issue
(If applicable, link to issue)

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Screenshots / Demo
N/A (dependency lockfile change)

## Checklist

- [x] My code follows the code style of this project
- [x] I have performed a self-review of my own changes
- [x] I have verified the fix locally
- [x] This change does not introduce new breaking changes
- [x] This change does not require documentation updates

## Additional Notes
This keeps the reproducibility benefits of `npm ci` while fixing the platform binding gap that caused Linux CI to fail.

Branch commit included in this PR:
- `3f5258e` — `ci: Avoid npm binding issues in SB deploy`

**Changed files:**
- `package-lock.json` (regenerated to include required optional platform bindings)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -83,14 +83,15 @@
       "license": "MIT"
     },
     "node_modules/@asamuzakjp/css-color": {
-      "version": "5.1.6",
-      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-5.1.6.tgz",
-      "integrity": "sha512-BXWCh8dHs9GOfpo/fWGDJtDmleta2VePN9rn6WQt3GjEbxzutVF4t0x2pmH+7dbMCLtuv3MlwqRsAuxlzFXqFg==",
+      "version": "5.1.11",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-5.1.11.tgz",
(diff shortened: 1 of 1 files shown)


andrewgos:
As discussed, let's try the option of regenerating the lockfile. Could be that we just happened to have dependencies inconsistencies, as we shouldn't need to regenerate every time. Because moving from `npm ci` to `npm install` sounds like a compromise on CI environment isolation.

Leaving it back to you to explore that option

andrewgos (approved):
❤️

PR #205, opened 2026-05-05
feat(activity-icon): Add new ActivityIcon component

andrewgos:
## Description
Adds the ActivityIcon component, a reusable icon renderer for Moodle activity and resource types.

It renders an icon from a central registry, supports `none | default | large` variants and `sm | md | lg | xl` sizes, and falls back safely when values are invalid (`assignment`, `default`, and `md`). It forwards native `HTMLAttributes<HTMLSpanElement>` props to the root element for flexibility.

Key implementation decisions:

- Uses one registry (`activityIconRegistry`) as the source of truth for icon assets and categories.
- Normalizes icon input and validates `icon`, `variant`, and `size` at runtime with safe defaults.
- Uses stable classes for variant, size, and category (`mds-activity-icon--*`) for predictable styling and tests.
- Uses Moodle design tokens for sizing, spacing, radius, and category backgrounds (no hardcoded design values).
- Keeps `none` variant icon-only, while `default` and `large` variants apply padded container styles.
- Supports accessibility via the `alt` prop (defaults to empty string when decorative).

https://moodle.atlassian.net/browse/MDS-543

## Screenshots/Previews
Screenshot 2026-05-05 at 10 58 47 am

### TypeScript and props

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity


components/activity-icon/activity-icon.css
Chocolate-lightning, line 1: Could we switch icon fills to currentColor (or document fixed-color intent)? Right now literal fills in the SVGs bypass the existing --mds-activity-icon-*-icon token pipeline and prevent theme/override support.

components/activity-icon/activityIconRegistry.ts
```diff
@@ -0,0 +1,133 @@
...
+  'file-graphic': { src: fileGraphicIcon, category: 'resource' },
+  'file-h5p': { src: fileH5pIcon, category: 'resource' },
+  'file-image': { src: fileImageIcon, category: 'resource' },
+  'file-isf-flowchart': { src: fileIsfFlowchartIcon, category: 'resource' },
+  'file-json': { src: fileJsonIcon, category: 'resource' },
+  'file-math': { src: fileMathIcon, category: 'resource' },
+  'file-moodle': { src: fileMoodleIcon, category: 'resource' },
+  'file-oth': { src: fileOthIcon, category: 'resource' },
+  'file-pdf': { src: filePdfIcon, category: 'resource' },
+  'file-plain text': { src: filePlainTextIcon, category: 'resource' },
```
Chocolate-lightning: Can we normalize this key to kebab-case (file-plain-text) to match the rest of the API surface? A space in the key is inconsistent with adjacent keys.

components/activity-icon/assets/bigbluebutton.svg
Chocolate-lightning, line 1: This SVG wraps an embedded 1200×1200 PNG data URI. Could we optimize this asset (or replace with vector) to avoid shipping a disproportionately large icon compared with the rest of the 24×24 set?

components/activity-icon/activity-icon.css
```diff
@@ -0,0 +1,96 @@
...
+  background-color: var(--mds-activity-icon-resource-bg);
+}
+
+/* Size variants */
+.mds-activity-icon--size-sm {
+  block-size: var(--mds-icons-sm);
+  inline-size: var(--mds-icons-sm);
+}
+
+.mds-activity-icon--size-md {
```
Chocolate-lightning, line 76: size=md currently resolves using --mds-icons-lg while --mds-icons-md is unused. If intentional (to match Figma optical sizing), could we add an inline comment clarifying this mapping?

components/activity-icon/ActivityIcon.tsx
```diff
@@ -0,0 +1,92 @@
...
+const allowedVariants: ActivityIconVariant[] = ['none', 'default', 'large'];
+
+export type ActivityIconSize = 'sm' | 'md' | 'lg' | 'xl';
+const allowedSizes: ActivityIconSize[] = ['sm', 'md', 'lg', 'xl'];
+
+export interface ActivityIconProps extends HTMLAttributes<HTMLSpanElement> {
+  /**
+   * Activity/resource icon key used to resolve the SVG asset from the registry.
+   */
+  icon: string;
```
Chocolate-lightning: Would it be safer to type icon as ActivityIconName or ActivityIconName | string ? Using string means invalid keys are caught only at runtime.

components/activity-icon/ActivityIcon.tsx
```diff
@@ -0,0 +1,92 @@
...
+    );
+  }
+
+  if (import.meta.env.DEV && size && !allowedSizes.includes(size)) {
+    console.warn(
+      `[MDS ActivityIcon] Invalid size "${size}". Falling back to "md". Allowed: ${allowedSizes.join(', ')}`,
+    );
+  }
+
+  const resolvedIcon: ActivityIconName = hasValidIcon
```
Chocolate-lightning, line 86: When icon lookup fails, we silently render assignment. Should we render nothing, a neutral placeholder or maybe fail entirely instead so integration mistakes remain visible in non-dev builds?

components/activity-icon/activityIconRegistry.ts
```diff
@@ -0,0 +1,133 @@
...
+  | 'other'
+  | 'resource';
+
+export interface ActivityIconRegistryEntry {
+  src: string;
+  category: ActivityIconCategory;
+}
+
+// Single source of truth: adding or removing an icon requires updating one entry here.
+export const activityIconRegistry = {
```
Chocolate-lightning: The registry eagerly imports all icon assets. Is this payload tradeoff intentional for this component, or should we consider a split/lazy strategy so single-icon usage does not pull the full set?

components/activity-icon/ActivityIcon.stories.tsx
```diff
@@ -0,0 +1,271 @@
...
+  activityIconRegistry,
+} from './activityIconRegistry';
+
+const meta = {
+  title: 'Components/ActivityIcon',
+  component: ActivityIcon,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'beta'],
```
Chocolate-lightning: Update this to stable.

Chocolate-lightning (changes requested):
Overall, great work on this component. I have a few targeted thoughts about the current implementation and some more general thoughts for potential future improvements:

1. Should we adopt a repo-wide lowercase kebab-case asset naming policy? Mixed-case/space filenames are brittle for scripts and globs.
2. Would a shared warnDev() helper be useful across components to avoid repeated environment checks/prefix formatting?
3. Should we define a shared size type in a central types module to avoid divergence as more components adopt sizing i.e. [sm, md, lg]?

components/activity-icon/activityIconRegistry.ts
andrewgos: nice spotting! I missed it while the AI was doing some transformations

components/activity-icon/activity-icon.css
andrewgos: good call. yes, it was intentional done in the figma (not only matching optical sizing but it's using the same token mapping), i'll add comments.

components/activity-icon/assets/bigbluebutton.svg
andrewgos: replaced with a 3KB svg file from Boris

components/activity-icon/ActivityIcon.tsx
andrewgos: good catch! forgot that we have that type ActivityIconName

components/activity-icon/activityIconRegistry.ts
andrewgos: it wasnt intentional, i'll work on a way to get it to lazy load 👍

components/activity-icon/ActivityIcon.tsx
andrewgos: After considering what the AI recommends, I'd go with the approach of a neutral placholder "file-unknown" icon and doing a console error that gives warns everyone (both developers and users) about the issue.

Let me know if you think it should be otherwise

components/activity-icon/activity-icon.css
andrewgos: i'd need more context as I don't quite understand your comment here 😬, are you proposing to alter the SVG content slightly then?

andrewgos:
Thanks @Chocolate-lightning for the lovely review!

I've actioned most of the items and replied to each inline comments, feel free to take a look and let me know if anything else is missed.

In regards to the general improvements, I would action that on a separate ticket/PR as it will involve changes that are repo-wide, perhaps for the next maintenance sprint?
- _Should we adopt a repo-wide lowercase kebab-case asset naming policy? Mixed-case/space filenames are brittle for scripts and globs._
  - Agree, would need to think the best tooling/check to verify these. The mixing of filenames in this PR was more of an accidental error 😬 Added to the backlog https://moodle.atlassian.net/browse/MDS-577
- _Would a shared warnDev() helper be useful across components to avoid repeated environment checks/prefix formatting?_
  - Yes, I think this would be helpful, making the component logic cleaner without worrying about environment checks. Added to the backlog https://moodle.atlassian.net/browse/MDS-576
- _Should we define a shared size type in a central types module to avoid divergence as more components adopt sizing i.e. [sm, md, lg]?_
  - Sorry I don't quite understand, how about those components that have more sizes than the basic three sm/md/lg?

Chocolate-lightning (approved):
Hey Andrew,

Thanks for addressing my feedback. I think for the fill suggestion, we can hold off for the time being until we have a need for different theming.

I see CI tests are good and testing it all locally also looks good.

Make sure to complete the MSE for the component and once that is done, I'll review that and merge this in and release a new version of the MDS.

components/activity-icon/activity-icon.css
andrewgos: resolving comment. this will be revisited when we have a requirement for theming

PR #216, opened 2026-05-12
feat(Checkbox): Add Checkbox component

Chocolate-lightning:
## Description

Adds the new `Checkbox` component to the design system, including implementation, component-scoped styles, Storybook stories, unit tests, Code Connect mappings, and package exports.

This PR introduces a checkbox input that follows the existing form-control patterns in the repo and supports the main Moodle design requirements:

- visible or visually hidden labels with accessible naming fallbacks
- checked, unchecked, indeterminate, invalid, and disabled states
- supporting text and invalid feedback messaging
- forwarded refs for form-library and focus-management compatibility
- RTL-safe layout and Storybook coverage
- Figma Code Connect mappings for the supported design variants

https://moodle.atlassian.net/browse/MDS-494

## Screenshots/Previews

- Storybook stories added for default, checked, controlled checked, disabled, invalid, invalid with feedback, supporting text, indeterminate, required indicator, hidden label, and RTL.
- Chromatic / preview links to be added when the PR is opened.

- [x] All 5 required files present: `Checkbox.tsx`, `checkbox.css`, `Checkbox.test.tsx`, `Checkbox.stories.tsx`, `index.tsx`

### TypeScript and props

- [x] Props interface extends the correct React HTML element interface (`InputHTMLAttributes<HTMLInputElement>`)

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

### Validation & QA

- Validation run: `VITEST_STORYBOOK=false npx vitest run components/checkbox/Checkbox.test.tsx`
- Result: 38/38 tests passed.
- Figma audit status: **Production-Ready** (~95% Figma alignment) — all design tokens correctly applied, all colour states match Figma exactly, typography fully aligned, spacing matches design specs.
- Minor visual differences (all negligible): ~1–2px wrapper height variance, baseline alignment imperceptible, flex layout achieves same visual spacing as design.

### Accessibility & Semantic HTML

- accessible-name fallback for hidden labels via `aria-label` then `label`
- `aria-invalid`, `aria-describedby`, and `aria-checked="mixed"` support
- visible required indicator tied to the label
- ref forwarding to the native input element
- Focus ring uses outline-only pattern (no box-shadow) for consistent keyboard navigation

### Design System Consistency

- Checkbox and Radio components now share identical layout and styling patterns (`inline-grid` siblings, sibling selectors for state, outline-based focus rings, RTL via logical properties)
- All CSS values use canonical `--mds-*` tokens; no legacy or non-standard token names
- Features full parity with Radio: label click-to-toggle via `htmlFor`, props spread for full attribute support, focus ring colour context (blue for default, red for invalid)

### DIFF
--- a/components/checkbox/Checkbox.stories.tsx
+++ b/components/checkbox/Checkbox.stories.tsx
@@ -1,0 +1,327 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { expect, waitFor } from 'storybook/test';
+import { Checkbox } from './Checkbox';
+
+const meta = {
+  title: 'Components/Checkbox',
+  component: Checkbox,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    label: {
+      description:
+        'Visible label text. Also used as the aria-label fallback when hideLabel is true and no explicit aria-label is provided.',
+      table: {
+        defaultValue: { summary: '' },
+      },
+    },
+    hideLabel: {
+      control: { type: 'boolean' },
+      description:
+        'When true, hides the visible label element. Use only when a visible label would be redundant. Always provide an accessible name via aria-label or label prop.',
+      table: {
+        type: {
+          summary: 'true | false',
+        },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    ['aria-label']: {
+      control: { type: 'text' },
+      description:
+        'Accessible label for the input. Takes precedence over the label prop when hideLabel is true. Required when hideLabel is true and no label prop is provided.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+      },
+    },
+    checked: {
+      control: { type: 'boolean' },
+      description: 'Whether the checkbox input is checked in controlled mode.',
+      table: {
+        type: {
+          summary: 'true | false',
+        },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    indeterminate: {
+      control: { type: 'boolean' },
+      description:
+        'Renders the checkbox in a mixed state, typically for parent/select-all controls.',
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    disabled: {
+      control: { type: 'boolean' },
+      description:
+        'Prevents interaction with the checkbox and applies disabled styling.',
+      table: {
+        type: {
+          summary: 'true | false',
+        },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    invalid: {
+      control: { type: 'boolean' },
+      description:
+        'Marks the input as invalid. Sets aria-invalid on the input and applies danger styling to the border and label.',
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    required: {
+      control: { type: 'boolean' },
+      description:
+        'Marks the checkbox as required for form submission. Displays a visual required indicator beside the label.',
+      table: {
+        type: {
+          summary: 'true | false',
+        },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    invalidFeedback: {
+      description:
+        'Pre-translated error message shown below the label when the input is invalid. Requires invalid={true} to be set.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'undefined' },
+      },
+    },
+    supportingText: {
+      description:
+        'Optional helper text shown below the label when no invalid feedback is active.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'undefined' },
+      },
+    },
+    id: {
+      control: { type: 'text' },
+      description:
+        'ID for the input element, used to associate the label. If omitted, a unique ID is generated automatically.',
+      table: {
+        type: {
+          summary: 'auto-generated | string',
+        },
+        defaultValue: { summary: 'auto-generated' },
+      },
+    },
+    name: {
+      control: { type: 'text' },
+      description:
+        'Name attribute submitted with the form. Groups related checkboxes under the same key.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+      },
+    },
+    value: {
+      control: { type: 'text' },
+      description:
+        'Value submitted with the form when the checkbox is checked.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '' },
+      },
+    },
+    autoFocus: {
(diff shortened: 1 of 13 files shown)


components/checkbox/checkbox.css
```diff
@@ -0,0 +1,182 @@
+.mds-checkbox {
```
andrewgos, line 1: whats the rationale of having a fallback values?

I'm afraid the fallback may raise an inconsistency when a token gets updated but forgetting this value to be updated too.

components/checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,309 @@
...
+    },
+    className: {
+      description:
+        'Additional class names to apply to the checkbox wrapper div for custom styling.',
+      table: {
+        defaultValue: { summary: '' },
+      },
+    },
+  },
+  args: {
```
andrewgos, line 156: i'd like to propose of adding the boolean's default value here

```
    hideLabel: false,
    checked: false,
    indeterminate: false,
    disabled: false,
    invalid: false,
    required: false,
    autoFocus: false,
```

which i think brings a better user experience without needing to click "set boolean" to let it show. and it is also not interfering with the code examples.


Image

components/checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,309 @@
...
+      description:
+        'ID for the input element, used to associate the label. If omitted, a unique ID is generated automatically.',
+      table: {
+        type: {
+          summary: 'auto-generated | string',
+        },
+        defaultValue: { summary: 'auto-generated' },
+      },
+    },
+    name: {
```
andrewgos, line 117: - It is missing the type for aria-label, name, value, and classNames
- and perhaps we can set an example of what the value may be? (though the default is `''`)

Similar to how we did it in Radio :)

components/checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,309 @@
...
+    await userEvent.click(checkbox);
+    await expect(checkbox).not.toBeChecked();
+    // Return to a neutral visual baseline so screenshot QA is not biased by focus styling.
+    checkbox.blur();
+    await expect(checkbox).not.toHaveFocus();
+    await expect(canvas.getByText('Remember this setting')).toBeVisible();
+  },
+} satisfies Story;
+
+export const Checked: Story = {
```
andrewgos: Label here should be `DefaultChecked` ?

components/checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,309 @@
...
+  args: {
+    defaultChecked: true,
+  },
+  play: async ({ canvas }) => {
```
andrewgos, line 194: I think something is off with the interactivity in storybook, It is not showing as checked, and I cannot make it checked. Applies to other stories too

https://github.com/user-attachments/assets/5eed5c77-1fb9-4bd1-a4d2-0e29e5768a62

components/checkbox/checkbox.css
```diff
@@ -0,0 +1,180 @@
...
+  align-self: start;
+  min-height: var(--mds-icons-lg, 1.5rem);
+  padding: var(--mds-spacing-none, 0);
+  column-gap: var(--mds-spacing-xs, 0.5rem);
+  row-gap: var(--mds-spacing-xxs, 0.25rem);
+
+  font-family: var(--mds-font-family-base, Arial, sans-serif);
+  font-size: var(--mds-font-size-paragraph-default, 1rem);
+  font-weight: var(--mds-font-weight-regular, 400);
+  line-height: var(--mds-line-height-paragraph-small, 1.0875rem);
```
andrewgos: For checkbox (also radio), I'm seeing the line-height on figma to be on `xs` now. is that correct @borisbasque ?

(also to the supporting text)

components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <div className={classes.join(' ')}>
+        <input
+          className={[
+            'mds-checkbox-input',
+            'form-check-input',
+            isInvalid ? 'is-invalid' : '',
+          ]
+            .filter(Boolean)
+            .join(' ')}
+          type="checkbox"
```
andrewgos: this may need to be after the `{...inputProps}` otherwise it'll get overridden by the consumers.

may be worth adding a unit test to ensure the types didn't get overridden

components/checkbox/index.tsx
```diff
@@ -0,0 +1,3 @@
+import './checkbox.css';
```
andrewgos: Just like the other component, the subpath entrypoints shouldn't include the css

```
// CSS is aggregated via components/index.css and imported by the package root entry (index.ts).
// It is intentionally omitted here so per-component subpath imports remain side-effect-free for JS consumers.
```

andrewgos (changes requested):
Nice work! 💪  Nothing major, just a few minor thoughts and some tidying up for the storybook 📖 

And just a curiousity question: what's the word `base` trying to imply in `Add base Checkbox component`?

components/checkbox/Checkbox.stories.tsx
Chocolate-lightning: Agreed, Implemented as suggested.

components/checkbox/Checkbox.tsx
Chocolate-lightning: Agreed, Implemented as suggested.

components/checkbox/index.tsx
Chocolate-lightning: Great catch, implemented as suggested.

components/checkbox/Checkbox.stories.tsx
Chocolate-lightning: Good catch, Updated as suggested.

components/checkbox/checkbox.css
Chocolate-lightning: Good call, I have removed the fallback values in the Checkbox component and updated the docs & instruction files to guardrail this.

components/checkbox/checkbox.css
Chocolate-lightning: Moved to xs value as suggested.

components/checkbox/Checkbox.stories.tsx
Chocolate-lightning: I am unable to replicate this currently both locally & via the SB build generated via Chromatic.

components/checkbox/Checkbox.stories.tsx
Chocolate-lightning: Suggestion taken on board and implemented. Cheers!

Chocolate-lightning:
Points from review have been addressed and visuals have been signed off. Merging this in as Andrew is away.

PR #222, opened 2026-05-18
feat(build): Build CSS variables at run-time from SCSS

andrewnicols:
## Description

Moodle allows tokens to be updated by an administrator at run-time. This patch exposes the SCSS variables as CSS variables which are resolved at compile time as an optional supplement to the scss/_index.scss.

## Type of Change

- [x] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have updated, added, and run tests such as JSUnit, Storybook interactions, or fuzzing to prove my fix is effective or that my feature works
- [ ] Instruction files updated if this change introduces new patterns or anti-patterns

## Additional Notes

As discusseed with @Chocolate-lightning


tokens/scss/_index_css_vars.scss
Chocolate-lightning, line 1: I think maybe we could look at combing this file & the _index.legacy.scss files into a _index_moodle.scss file so that it contains any special handling needed for the core LMS to incorporate the MDS tokens in. Saves having segmented loose files that need to be imported on your side of the pond. Thoughts?

Chocolate-lightning (changes requested):
Kia ora @andrewnicols,

Welcome into the MDS project!

Looking at this PR and as we discussed previously, we can include such a change and I only have a few notes:

1. The question on the new index file
2. The commit should likely be "feat(Tokens): Build CSS variables at run-time from SCSS"
3. A quick rebase would be appreciated

Cheers,

tokens/scss/_index_css_vars.scss
andrewnicols: I disagree. The legacy file has a completely different purpose - it's only there because Moodle has an older version of the SCSS compiler which does not support `@forward`.

The CSS variables is intended to support Moodle plugins which do not have access to SCSS at all.

Once Moodle updates to a recent version of SCSS, which I believe is on the radar, the legacy file will not be included, but we will still need the CSS variable definitions from SCSS at run time.

andrewnicols:
Thanks @Chocolate-lightning,

1. I've added a comment, but I disagree and have not made a change here.
2. Fixed
3. Rebased

tokens/scss/_index_css_vars.scss
andrewnicols: Ah - just re-read this in a different light... rethinking.

Chocolate-lightning (approved):
Kia ora @andrewnicols,

Thanks for addressing the points raised during the review, Local tests are still good and the CI Chromatic is a non issue as we are not modifying actual component token usages.

Accepting this in.

PR #233, opened 2026-05-22
fix(Tokens): Darken info default color for a11y

borisbasque:


Chocolate-lightning (approved):
Change from ZH & Figma with the token files updated to reflect the new value.

This was updated due to finding that the set value did not quite meet WCAG defined values so it was darkened a little bit for compliance.

PR #241, opened 2026-06-03
feat(Pagination): Introduce pagination component

Chocolate-lightning:
## Description

This PR introduces the new `Pagination` component to the Moodle Design System with support for two rendering modes and state-aware interaction patterns:

- `full` mode for multi-page navigation with page numbers and ellipsis handling
- `grouped` mode for previous/next-only navigation when total-page context is intentionally omitted

The implementation follows the component checklist workflow and aligns naming/behaviour across code, Storybook, and Figma. The main goal is to provide a reusable, accessible pagination primitive that can be consumed consistently in Moodle interfaces while preserving token-driven styling and predictable keyboard/screen-reader behaviour.

### General approach

1. Design-first parity

- Use the Figma source as the behaviour and visual reference baseline: https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=8467-37488&m=dev
- Verify mode semantics (`full` vs `grouped`), control affordances, and state transitions (`default`, `hover`, `active`, `disabled`) against the design intent.

2. API and rendering model

- Provide a constrained, explicit prop API to avoid ambiguous pagination behaviour.
- Keep rendering logic deterministic for edge cases (small page counts, near-start/end windows, ellipsis placement).
- Ensure invalid/unsupported prop values gracefully fall back to documented defaults.

3. Styling and token compliance

- Use existing `--mds-*` tokens and preserve the project guardrail of no fallback literals where matching tokens exist.
- Keep variants composable via stable class hooks (`mds-*`) and predictable CSS structure.

4. Accessibility and interaction quality

- Maintain correct `aria-current` usage for active page in `full` mode.
- Ensure icon-only controls are labelled (`aria-label`) and disabled states are exposed correctly.
- Validate focus visibility and keyboard navigation behaviour in stories/tests.

5. Verification and rollout readiness

- Cover helper logic with focused unit tests.
- Cover visual/interactive states via Storybook stories and test tags.
- Use the MDS build/checklist guidance as a readiness gate before merge.

- Jira: https://moodle.atlassian.net/browse/MDS-496
- Confluence checklist/reference: https://moodle.atlassian.net/wiki/spaces/MDS/pages/3937632270/Pagination+MSE+and+Build+Checklist
- Figma design reference: https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=8467-37488&m=dev

### TypeScript and props

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes

- This draft intentionally follows the component PR template format so it can be copied into a GitHub PR opened with `?template=component.md`.
- Reviewer focus requested:
  - Parity sign-off for grouped/full mode behaviour against Figma node `8467:37488`
  - Validation against MDS-496 acceptance intent
  - Final pass against the Pagination MSE and Build Checklist before merge

### DIFF
--- a/.storybook/preview.css
+++ b/.storybook/preview.css
@@ -12,3 +12,49 @@
   font-weight: 100 900;
   font-style: italic;
 }
+
+/* ---------------------------------------------------------------------------
+ * Pseudo-state class equivalents for storybook-addon-pseudo-states.
+ *
+ * The addon rewrites document stylesheets at runtime to add .pseudo-* class
+ * selectors alongside every :pseudo-state rule. In Vite dev mode the CSS is
+ * injected into the iframe after the addon's STORY_RENDERED rewrite pass, so
+ * the rewrite never lands on our component stylesheets. Adding the equivalents
+ * here (a static file loaded before stories render) is the correct fallback.
+ *
+ * Rules mirror the `:focus-visible` declarations in each component's CSS file.
+ * Note: do NOT combine real pseudo-states (e.g. :active) with .pseudo-* class
+ * names in this file — the addon will attempt to rewrite such rules, produce
+ * an empty selector, and throw a CSSStyleSheet insertRule error.
+ * --------------------------------------------------------------------------- */
+
+/* All components — default focus ring */
+.mds-btn.btn.pseudo-focus-visible,
+.mds-checkbox-input.pseudo-focus-visible,
+.mds-form-check-input.pseudo-focus-visible,
+.mds-pagination__button.pseudo-focus-visible:not(:disabled),
+.mds-pagination__page.pseudo-focus-visible:not(:disabled) {
+  outline: var(--mds-stroke-weight-md) solid var(--mds-focus-default);
+  outline-offset: var(--mds-spacing-offset);
+  box-shadow: none;
+}
+
+/* Close button — different offset and opacity */
+.mds-close-button.btn-close.pseudo-focus-visible {
+  outline: var(--mds-stroke-weight-md) solid var(--mds-focus-default);
+  outline-offset: var(--mds-spacing-none);
+  box-shadow: none;
+  opacity: 1;
+}
+
+/* Button danger — outline color override */
+.mds-btn.btn.btn-danger.pseudo-focus-visible,
+.mds-btn.btn.btn-outline-danger.pseudo-focus-visible {
+  outline-color: var(--mds-focus-danger);
+}
+
+/* Checkbox / radio invalid — outline color override */
+.mds-checkbox-input.is-invalid.pseudo-focus-visible,
+.mds-form-check-input.is-invalid.pseudo-focus-visible {
+  outline-color: var(--mds-border-feedback-danger);
+}
--- a/components/pagination/Pagination.figma.tsx
+++ b/components/pagination/Pagination.figma.tsx
@@ -1,0 +1,61 @@
+import figma from '@figma/code-connect';
+import { Pagination } from './Pagination';
+
+// Use the Pagination component set node (not a docs frame instance) so Dev Mode
+// can apply this mapping consistently across all shown examples.
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=8367-5141';
+
+// Full variant: pagination with page numbers.
+figma.connect(Pagination, url, {
+  variant: { 'Current page': 'Middle page', 'Total pages': '9 or more' },
+  example: () => (
+    <Pagination
+      totalPages={10}
+      currentPage={5}
+      ariaLabel="Pagination"
+      onPageChange={() => {}}
+    />
+  ),
+});
+
+// Grouped variant: previous and next controls only.
+// In the current Figma component set this is represented by Total pages=None.
+figma.connect(Pagination, url, {
+  variant: { 'Current page': 'Middle page', 'Total pages': 'None' },
+  example: () => (
+    <Pagination
+      totalPages={10}
+      currentPage={5}
+      variant="grouped"
+      ariaLabel="Pagination"
+      onPageChange={() => {}}
+    />
+  ),
+});
+
+// First page variant: previous button disabled.
+figma.connect(Pagination, url, {
+  variant: { 'Current page': 'First page', 'Total pages': '9 or more' },
+  example: () => (
+    <Pagination
+      totalPages={10}
+      currentPage={1}
+      ariaLabel="Pagination"
+      onPageChange={() => {}}
+    />
+  ),
+});
+
+// Last page variant: next button disabled.
+figma.connect(Pagination, url, {
+  variant: { 'Current page': 'Last page', 'Total pages': '9 or more' },
+  example: () => (
+    <Pagination
+      totalPages={10}
+      currentPage={10}
+      ariaLabel="Pagination"
+      onPageChange={() => {}}
+    />
+  ),
+});
--- a/components/pagination/Pagination.stories.tsx
+++ b/components/pagination/Pagination.stories.tsx
@@ -1,0 +1,611 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { useState } from 'react';
+import { expect, userEvent } from 'storybook/test';
+import { Pagination } from './Pagination';
+
+const meta = {
+  title: 'Components/Pagination',
+  component: Pagination,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    totalPages: {
+      control: { type: 'number', min: 1, max: 1000, step: 1 },
+      description: 'Total number of pages',
+      table: {
+        type: { summary: 'number' },
+      },
+    },
+    currentPage: {
+      control: { type: 'number', min: 1, max: 1000, step: 1 },
+      description: 'Current page number (1-indexed)',
+      table: {
+        type: { summary: 'number' },
+      },
+    },
+    onPageChange: {
+      description: 'Callback fired when the page changes',
+      table: {
+        type: { summary: '(page: number) => void' },
+      },
+    },
+    ariaLabel: {
+      control: { type: 'text' },
+      description: 'Accessible name for the pagination landmark.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: "'Pagination'" },
+      },
+    },
+    previousPageLabel: {
+      control: { type: 'text' },
+      description: 'Accessible label used for the previous-page button.',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: "'Previous page'" },
+      },
+    },
+    nextPageLabel: {
+      control: { type: 'text' },
+      description: 'Accessible label used for the next-page button.',
+      table: {
+        type: { summary: 'string' },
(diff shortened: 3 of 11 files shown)


.storybook/preview.css
```diff
@@ -11,3 +11,49 @@
   font-weight: 100 900;
   font-style: italic;
 }
+
+/* ---------------------------------------------------------------------------
+ * Pseudo-state class equivalents for storybook-addon-pseudo-states.
```
andrewgos, line 17: trivial question:

Would having this cause inconsistency when there are changes to the focus ring? Because this is a separate source to maintain anytime there are changes happening to the code. For example, thinking the look was fine on the add-on while it was actually different from the real focus-visible run.

I wonder if there's a way to sync this up or a warning whenever there's a difference? 🤔

components/pagination/Pagination.figma.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Pagination } from './Pagination';
+
+// Use the Pagination component set node (not a docs frame instance) so Dev Mode
+// can apply this mapping consistently across all shown examples.
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=8367-5141';
+
+// Full variant: pagination with page numbers.
+figma.connect(Pagination, url, {
+  variant: { 'Current page': 'middle page', 'Total pages': '9 or more' },
```
andrewgos: probably one to bring up to the designer:

Image

why one is not capitalised `m` 🥲

components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,500 @@
...
+      },
+    },
+    variant: {
+      control: { type: 'select' },
+      options: ['full', 'grouped'],
+      description:
+        "Controls which variant to render. `'full'` shows page numbers (default). `'grouped'` shows only previous and next controls. When `'full'`, the component automatically collapses to grouped appearance on very narrow viewports.",
+      table: {
+        type: { summary: "'full' | 'grouped'" },
+        defaultValue: { summary: "'full'" },
```
andrewgos, line 74: this one is a select, probably won't need to display the quotes

```
        type: { summary: 'full | grouped' },
        defaultValue: { summary: 'full' },
```

components/pagination/Pagination.figma.tsx
```diff
@@ -0,0 +1,35 @@
+import figma from '@figma/code-connect';
+import { Pagination } from './Pagination';
+
+// Use the Pagination component set node (not a docs frame instance) so Dev Mode
+// can apply this mapping consistently across all shown examples.
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=8367-5141';
+
+// Full variant: pagination with page numbers.
```
andrewgos, line 9: would we want more for the full variants with 'first page' and 'last page' too ?

components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,500 @@
...
+const meta = {
+  title: 'Components/Pagination',
+  component: Pagination,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    totalPages: {
+      control: { type: 'number', min: 1, max: 100, step: 1 },
```
andrewgos: i think the max on totalPages and currentPage here might be misleading/limiting. some users might want to test the looks when they go thousands

components/pagination/pagination.css
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/* Shared base styles for buttons and page items */
+.mds-pagination__button,
+.mds-pagination__page {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-width: 36px;
+  min-height: 36px;
```
andrewgos: Looks like theres differences on the height being 34px in Figma for page numbers, ellipsis, and controls.

components/pagination/pagination.css
```diff
@@ -0,0 +1,130 @@
...
+  color: var(--mds-text-muted);
+  cursor: not-allowed;
+}
+
+/* Icon sizing */
+.mds-pagination__button i {
+  /* width/font-size use --mds-icons-sm (= scale-600 = 1rem = 16px, the icon md size). */
+  width: var(--mds-icons-sm);
+  font-size: var(--mds-icons-sm);
+  line-height: var(--mds-line-height-paragraph-small);
```
andrewgos: also line-heights in figma is xs instead of small

components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,500 @@
...
+    await expect(
+      canvas.getByRole('button', { name: 'Go to previous results page' }),
+    ).toBeInTheDocument();
+    await expect(
+      canvas.getByRole('button', { name: 'Results page 4' }),
+    ).toBeInTheDocument();
+  },
+};
+
+export const PseudoStateMatrix: Story = {
```
andrewgos, line 320: I think we can do some styling rework on this story, the alignment is squeezed and off

Image

or perhaps we just show a single page item or a single chevron for each? (similar to how we have it on Figma) instead of showing the entire pagination bar

components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,500 @@
...
+  ),
+  play: async ({ canvas }) => {
+    const disabledControls = canvas
+      .getAllByRole('button')
+      .filter((button) => button.hasAttribute('disabled'));
+    await expect(disabledControls.length).toBeGreaterThan(0);
+  },
+};
+
+export const PageDistributionAndBoundaryValidationBundle: Story = {
```
andrewgos, line 368: Also the styling here can also be improved?

Image

maybe label on the left with the output on the right? something like this:

Image

components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const classes = ['mds-pagination', `mds-pagination--${effectiveVariant}`];
+  if (className) {
+    classes.push(className);
+  }
+
+  // First, middle, and last page buttons share identical behavior; keeping the
+  // markup here avoids three nearly identical JSX branches below.
+  const renderPageButton = (page: number) => (
+    <button
```
andrewgos, line 158: i might be wrong, but i'm guessing a bare button might accidentally submit a surrounding form if theres any

We can add `type="button"` to be safe

andrewgos: maybe also a unit test?

```
  it('does not submit a surrounding form when a pagination button is clicked', async () => {
    const handleSubmit = vi.fn((event: React.FormEvent<HTMLFormElement>) => {
      event.preventDefault();
    });

    render(
      <form onSubmit={handleSubmit}>
        <Pagination {...defaultProps} currentPage={2} />
      </form>,
    );

    await userEvent.click(screen.getByRole('button', { name: /page 3/i }));

    expect(handleSubmit).not.toHaveBeenCalled();
  });

```

andrewgos (changes requested):
Nicely done with the dynamic ellipsis and viewport limits 🤩

I've got several items that I've noticed needs fixing and also some styling improvements idea that can be done 😉

components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,194 @@
+import type { ComponentPropsWithoutRef } from 'react';
+import { useMemo } from 'react';
+import './pagination.css';
```
andrewgos: One more thing: we would need to remove this line as it will causes the built to have `Pagination2.js`

components/pagination/Pagination.stories.tsx
Chocolate-lightning: Done. I updated the summary to remove quoted literals for the select display:

`type: { summary: 'full | grouped' }`

This now matches Storybook control semantics more cleanly.

components/pagination/Pagination.figma.tsx
Chocolate-lightning: Done. I added explicit Code Connect entries for:

- `Current page: first page`, `Total pages: 9 or more`
- `Current page: last page`, `Total pages: 9 or more`

This complements the existing middle-page full/grouped mappings and closes the boundary-state gap.

components/pagination/Pagination.stories.tsx
Chocolate-lightning: Done. I increased the controls for both `currentPage` and `totalPages` from 100 to 1000 so reviewers can test high-page-count visuals and behavior.

components/pagination/pagination.css
Chocolate-lightning: Done. I aligned the sizing to 34px for controls/page items/ellipsis to match Figma parity.

components/pagination/pagination.css
Chocolate-lightning: Done. I switched the relevant pagination icon/text line-height usage to the xs token (`--mds-line-height-paragraph-xs`) per Figma.

components/pagination/Pagination.stories.tsx
Chocolate-lightning: Done. I reworked the matrix layout to improve readability and spacing (wider label/state columns and less cramped alignment), while still keeping the state comparison structure intact.

If you’d still prefer the more minimal “single control per state cell” presentation (closer to Figma), I can do a quick follow-up simplification.

components/pagination/Pagination.tsx
Chocolate-lightning: Done.

components/pagination/Pagination.tsx
Chocolate-lightning: Done on both points:

- added `type="button"` to pagination buttons to avoid accidental form submission,
- added a unit test that renders Pagination inside a form and verifies click interactions do not trigger submit.

.storybook/preview.css
Chocolate-lightning: Thanks, this is a good callout. The fallback in `.storybook/preview.css` is only to bridge the pseudo-state addon rewrite gap in Vite dev mode, but I agree it introduces duplication risk against component-level `:focus-visible` rules.

I’m happy to follow up with one of these options in a separate hardening pass:

- move these Storybook pseudo focus rules into a shared generated source that mirrors component tokens,
- add a visual regression/check that fails when pseudo fallback and component focus ring diverge,
- or replace the addon path entirely if we can get a stable Storybook-native state simulation without mirrored CSS.

For this PR, I kept parity with current component focus token values so visual behavior remains aligned.

components/pagination/Pagination.figma.tsx
Chocolate-lightning: great spot, actioned.

.storybook/preview.css
andrewgos: Sounds good to me. In the meantime i've created a ticket for the backlog https://moodle.atlassian.net/browse/MDS-597

andrewgos:
Thanks for actioning it all! Everything looks good to me now. Happy to pass the styling of the matrix too 👍 

In regards to the hardening of pseudo-state consistency between Storybook and real component rules, I've create a separate ticket in the backlog https://moodle.atlassian.net/browse/MDS-597

Pinging @borisbasque for the Chromatic review.

andrewgos (approved):

PR #243, opened 2026-06-04
feat(ProgressBar): Add new ProgressBar component

andrewgos:
## Description

This PR introduces a new ProgressBar component with Moodle Design System token-based styling, Storybook coverage, unit tests, and Figma Code Connect mapping.

**Key implementations**

- Implemented safe runtime handling: clamps value to 0-100, falls back to invalid status/label variant to defaults, and logs dev warnings for invalid inputs.
- Added visual state override rules so 0% always renders as empty and 100% always renders as completed, regardless of status prop.
- Implemented accessibility semantics on the progress track with role progressbar, aria value attributes, and robust accessible naming via aria-labelledby or aria-label fallback.

https://moodle.atlassian.net/browse/MDS-532

### TypeScript and props

### Internationalisation

### CSS and tokens

  - There lives some hardcoded values, see additional notes below

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes
- The height and min-width of the track line intentionally have no tokens (confirmed by the designers) to provide flexibility

### DIFF
--- a/components/progress-bar/ProgressBar.stories.tsx
+++ b/components/progress-bar/ProgressBar.stories.tsx
@@ -1,0 +1,315 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { expect } from 'storybook/test';
+import { ProgressBar } from './ProgressBar';
+
+const meta = {
+  title: 'Components/ProgressBar',
+  component: ProgressBar,
+  parameters: {
+    layout: 'padded',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    value: {
+      control: { type: 'number' },
+      description:
+        'Current progress value in the range defined by min and max. Fill is normalised from min to max.',
+      table: {
+        type: { summary: 'number' },
+        defaultValue: { summary: '0' },
+      },
+    },
+    min: {
+      control: { type: 'number' },
+      description: 'Lower bound for the progress range.',
+      table: {
+        type: { summary: 'number' },
+        defaultValue: { summary: '0' },
+      },
+    },
+    max: {
+      control: { type: 'number' },
+      description: 'Upper bound for the progress range.',
+      table: {
+        type: { summary: 'number' },
+        defaultValue: { summary: '100' },
+      },
+    },
+    status: {
+      control: { type: 'select' },
+      options: ['in-progress', 'loading', 'error', 'warning'],
+      description:
+        'Visual state — controls fill and track colours only when value is between 1 and 99.',
+      table: {
+        type: {
+          summary: 'in-progress | loading | error | warning',
+        },
+        defaultValue: { summary: 'in-progress' },
+      },
+    },
+    animated: {
+      control: { type: 'boolean' },
+      if: { arg: 'status', eq: 'loading' },
+      description:
+        'Controls striped animation when status is loading. Ignored for non-loading statuses.',
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    labelVariant: {
+      control: { type: 'select' },
+      options: ['title-and-count', 'title', 'inline', 'none'],
+      description: 'Controls label layout.',
+      table: {
+        type: { summary: 'title-and-count | title | inline | none' },
+        defaultValue: { summary: 'title-and-count' },
+      },
+    },
+    title: {
+      description:
+        'Pre-translated title text rendered above the bar, or used as the accessible name when no visible label is present.',
+      table: { defaultValue: { summary: '' } },
+    },
+    count: {
+      description:
+        'Pre-translated count or percentage text, e.g. "3 of 10" or "50%". The consuming app owns this string so locale-specific word order, plural rules, and number formatting can come from its i18n layer.',
+      table: { defaultValue: { summary: '' } },
+    },
+  },
+  args: {
+    value: 50,
+    min: 0,
+    max: 100,
+    status: undefined,
+    animated: false,
+    labelVariant: undefined,
+    title: 'Uploading files',
+    count: '5 of 10',
+  },
+  play: async ({ canvas }) => {
+    await expect(canvas.getByRole('progressbar')).toBeVisible();
+  },
+} satisfies Meta<typeof ProgressBar>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+const showcaseParameters = {
+  controls: { disable: true },
+  docs: {
+    canvas: { sourceState: 'none' },
+  },
+} as const;
+
+const showcaseGridStyle = {
+  display: 'grid' as const,
+  gap: 'var(--mds-spacing-md)',
+};
+
+const showcaseRowStyle = {
+  display: 'grid' as const,
+  gridTemplateColumns: 'minmax(10rem, 12rem) minmax(0, 1fr)',
+  gap: 'var(--mds-spacing-md)',
+  alignItems: 'center' as const,
+  minHeight: 'var(--mds-spacing-xxl)',
+};
+
+const showcaseLabelStyle = {
+  color: 'var(--mds-text-subtle)',
+  fontSize: 'var(--mds-font-size-paragraph-small)',
+  fontFamily: 'var(--mds-font-family-base)',
+  fontWeight: 'var(--mds-font-weight-medium)',
+};
+
+export const Default: Story = {};
+
+export const LabelVariants: Story = {
+  parameters: showcaseParameters,
+  play: async ({ canvas }) => {
+    await expect(canvas.getAllByRole('progressbar')).toHaveLength(4);
+  },
+  render: () => (
+    <div style={showcaseGridStyle}>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>Title and count</span>
+        <ProgressBar
+          value={50}
+          status="in-progress"
+          labelVariant="title-and-count"
+          title="Uploading files"
+          count="5 of 10"
+        />
+      </div>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>Title only</span>
+        <ProgressBar
+          value={50}
+          status="in-progress"
+          labelVariant="title"
+          title="Uploading files"
+        />
+      </div>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>Inline count</span>
+        <ProgressBar
+          value={50}
+          status="in-progress"
+          labelVariant="inline"
+          title="Uploading files"
+          count="5 of 10"
+        />
+      </div>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>No visible label</span>
+        <ProgressBar
+          value={50}
+          status="in-progress"
+          labelVariant="none"
+          title="Uploading files"
+        />
+      </div>
+    </div>
+  ),
+};
+
+export const Statuses: Story = {
+  parameters: showcaseParameters,
+  play: async ({ canvas }) => {
+    await expect(canvas.getAllByRole('progressbar')).toHaveLength(6);
+  },
+  render: () => (
+    <div style={showcaseGridStyle}>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>Empty (0%)</span>
+        <ProgressBar
+          value={0}
+          status="warning"
+          labelVariant="title-and-count"
+          title="Not started"
+          count="0 of 10"
+        />
+      </div>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>In progress</span>
+        <ProgressBar
+          value={50}
+          status="in-progress"
+          labelVariant="title-and-count"
+          title="Uploading files"
+          count="5 of 10"
+        />
+      </div>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>Loading</span>
+        <ProgressBar
+          value={30}
+          status="loading"
+          labelVariant="title-and-count"
+          title="Loading"
+          count="3 of 10"
+        />
+      </div>
+      <div style={showcaseRowStyle}>
+        <span style={showcaseLabelStyle}>Error</span>
+        <ProgressBar
+          value={40}
+          status="error"
+          labelVariant="title-and-count"
+          title="Upload failed"
+          count="4 of 10"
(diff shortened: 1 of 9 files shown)


components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,175 @@
...
+      console.warn(
+        `[MDS ProgressBar] value "${value}" is outside the allowed range [0–100] and will be clamped.`,
+      );
+    }
+    // Warn when no accessible name can be derived for the progressbar track.
+    const noVisibleTitle =
+      resolvedLabel === 'inline' || resolvedLabel === 'none';
+    const hasExplicitAriaLabel =
+      'aria-label' in props || 'aria-labelledby' in props;
+    if (noVisibleTitle && !title && !hasExplicitAriaLabel) {
```
Chocolate-lightning: The component only warns in development when labelVariant is inline or none and no accessible name is provided, but still renders <div role="progressbar"> without a stable accessible name in production. This is an a11y blocker because assistive tech can encounter an unnamed progress indicator.

Maybe we should look at enforcing a aria-describedby or enforcing an accessible name at runtime and development to avoid potential issues.

components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,175 @@
...
+  );
+  const isValueInRange = value >= 0 && value <= 100;
+
+  const resolvedStatus: ProgressBarStatus = isAllowedStatus
+    ? status
+    : 'in-progress';
+  const resolvedLabel: ProgressBarLabelVariant = isAllowedLabelVariant
+    ? labelVariant
+    : 'title-and-count';
+  const clampedValue = Math.min(100, Math.max(0, value));
```
Chocolate-lightning: aria-valuemin/aria-valuemax are fixed to 0/100, and value is forcibly clamped to 0..100. This limits compatibility with real-world progress sources e.g. (3 of 10, 230 of 500)

Perhaps we could add min/max props (default 0/100), clamp against that range, and derive percentage for visual width separately?

components/progress-bar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,274 @@
...
+      description:
+        'Pre-translated title text rendered above the bar, or used as the accessible name when no visible label is present.',
+      table: { defaultValue: { summary: '' } },
+    },
+    count: {
+      description:
+        'Pre-translated count or percentage text, e.g. "3 of 10" or "50%".',
+      table: { defaultValue: { summary: '' } },
+    },
+    animated: {
```
Chocolate-lightning: In docs/controls, animated is always visible even though it is only effective in loading state. This creates confusion feels odd when updating the value and not seeing any perceptible changes. We can likely add a conditional for this arg like `if: { arg: 'status', eq: 'loading' },` as noted in https://storybook.js.org/docs/api/arg-types#if

components/progress-bar/progress-bar.css
```diff
@@ -0,0 +1,113 @@
...
+
+/* Title text */
+.mds-progress-bar-title {
+  font-weight: var(--mds-font-weight-medium);
+  line-height: var(--mds-line-height-paragraph-small);
+  word-break: break-word;
+}
+
+/* Count / percentage text */
+.mds-progress-bar-count {
```
Chocolate-lightning, line 167: `count` uses `white-space: nowrap` with `flex-shrink: 0`, so long localized strings can force horizontal overflow on small screens (observed at 320px viewport in Storybook runtime checks).

components/progress-bar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,274 @@
...
+          labelVariant="title-and-count"
+          title="Loading"
+          count="3 of 10"
+        />
+      </div>
+    </div>
+  ),
+};
+
+export const RightToLeft: Story = {
```
Chocolate-lightning, line 281: No RTL story section demonstrating animated loading behavior. This leaves a gap for the checklist item around directional behavior under loading/animation.

components/progress-bar/ProgressBar.test.tsx
```diff
@@ -0,0 +1,375 @@
...
+    render(<ProgressBar title="Upload" value={50} />);
+    expect(screen.getByRole('progressbar')).toBeInTheDocument();
+  });
+
+  it('uses in-progress visual styling when status is omitted and value is between 1 and 99', () => {
+    const { container } = render(<ProgressBar title="Upload" value={50} />);
+    expect(container.firstChild).toHaveClass('mds-progress-bar--in-progress');
+  });
+
+  it('sets aria-valuenow, aria-valuemin, and aria-valuemax on the track', () => {
```
Chocolate-lightning, line 23: Current tests assert static ARIA values but do not explicitly verify value updates across rerenders.

Maybe we should add a rerender test asserting aria-valuenow and style.width update as props change, and that status transitions remain consistent with override rules?

components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,175 @@
...
+   * - `none`: bar only, no visible label
+   */
+  labelVariant?: ProgressBarLabelVariant;
+  /** Pre-translated title text rendered above the bar (title-and-count / title variants).
+   *  Also used as the accessible name for the bar when no visible label is present. */
+  title?: string;
+  /** Pre-translated count or percentage text, e.g. "3 of 10" or "50%". */
+  count?: string;
+  /** Controls striped animation when status is loading. */
+  animated?: boolean;
```
Chocolate-lightning, line 38: More of a general question for the designers I suppose but, Bootstrap 5 documents animated stripes moving right-to-left. This is reflected in the current implementation but I am wondering if we should purposefully diverge here and treat the animation as more of a "filling up the bar" rather than the current directionality?

components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,175 @@
...
+   * - `title`: title only above the bar
+   * - `inline`: bar with count beside it in a row
+   * - `none`: bar only, no visible label
+   */
+  labelVariant?: ProgressBarLabelVariant;
+  /** Pre-translated title text rendered above the bar (title-and-count / title variants).
+   *  Also used as the accessible name for the bar when no visible label is present. */
+  title?: string;
+  /** Pre-translated count or percentage text, e.g. "3 of 10" or "50%". */
+  count?: string;
```
Chocolate-lightning, line 36: I had the thought when doing the initial review that if we could move from caller-owned count text. That pattern is valid for i18n flexibility, but perhaps we could look at providing structured formatting helpers e.g. ({current} {of} {total}). We could also look at deriving this value based on the values given to us and having it calculated internally instead.

Keen to hear your thoughts on this.

components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+type ProgressBarVisualStatus = 'empty' | ProgressBarStatus | 'completed';
+
+type ProgressBarLabelVariant = 'title-and-count' | 'title' | 'inline' | 'none';
+
+export interface ProgressBarProps extends HTMLAttributes<HTMLDivElement> {
+  /** Progress value from 0 to 100. Values outside this range are clamped. */
+  value?: number;
+  /** Visual state of the progress indicator — controls fill and track colours. */
+  status?: ProgressBarStatus;
```
Chocolate-lightning, line 18: The component uses runtime fallback for `status`/`labelVariant`, but the props are still typed as narrow unions. The project instruction expects constrained props to accept broad runtime input (e.g., `string`) so JS consumers are explicitly supported without type assertions and so runtime guards are the source of truth. This is an instruction-file deviation.

Perhaps we should either align with `.github/instructions/components.instructions.md` (“Runtime validation for constrained props”): widen external prop type to `string` for constrained props, keep `allowed*` arrays as canonical runtime validators, and resolve to safe defaults with dev warnings. The other option is to reevaluate that agent rule in the instruction file to better reflect our stance on props and typing validation.

components/progress-bar/progress-bar.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/* Inline variant: bar on the left, count on the right in one row. */
+.mds-progress-bar.mds-progress-bar--label-inline {
+  flex-direction: row;
+  align-items: center;
+}
+
+/* Track (outer container) */
+.mds-progress-bar-track.progress {
```
Chocolate-lightning, line 18: The component intentionally uses Bootstrap classes (`progress`, `progress-bar`, `progress-bar-striped`, `progress-bar-animated`) but does not fully isolate from host-level Bootstrap overrides.

In Moodle installs that override `.progress` or `--bs-progress-*` globally, MDS rendering can drift (height, stripe sizing, transition timing, etc.). We should likely hardened this before merge.

To address this, we should look at explicitly setting the relevant Bootstrap progress custom properties on the MDS track/fill scope so host overrides do not leak unexpectedly i.e. `--bs-progress-height`

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

The ProgressBar implementation is close to merge-ready and includes good baseline coverage: component wiring, Storybook scenarios, unit tests, and Figma Code Connect are all present. The main blockers are around accessibility hardening for unnamed progressbars and Bootstrap 5 coexistence hardening in environments that globally customize `.progress`/`.progress-bar`. There are also several non-blocking but meaningful improvements around API flexibility (`min`/`max` semantics), Storybook control UX, RTL/animation coverage, and responsive behavior for long `count` strings.

components/progress-bar/progress-bar.css
andrewgos: Added attributes to our progress related classes 👍

components/progress-bar/progress-bar.css
andrewgos: updated so that now title and count has its boundaries for its reponsiveness on both `title-and-count` and `inline` variant

components/progress-bar/ProgressBar.stories.tsx
andrewgos: good call! applied this for the `animated`

unfortunately we cant use it for the `title` and `count`,  as its not possible to do multiple conditional ifs 

I can see this as retrospective improvement for our earlier storybooks too. ie. Radio's `invalidFeedback` only when `invalid` is `true`.

components/progress-bar/ProgressBar.stories.tsx
andrewgos: good catch! found that the animation is wrong.

fixed the direction of the animation and also the slanted angle to be mirrored too

components/progress-bar/ProgressBar.test.tsx
andrewgos: test added for the updates on rerenders 👍

components/progress-bar/ProgressBar.tsx
andrewgos: As far as I can see, other components are also doing constrained props in the types. 

Updated the code with option number 2, that is to update the instructions files ✅

components/progress-bar/ProgressBar.tsx
andrewgos: I think the biggest concern will be on the language, I'm afraid with that structured format where `of` will be restricted to English, which is unfair for the other languages. From the perspective of the component itself, it's on a scope that won't be able to discern what are the language is running in the larger scope/parents.

wdyt?

components/progress-bar/ProgressBar.tsx
andrewgos: Checked with Boris, we should just follow what Boostrap 5 has (that is its slanted angle and its direction)

Additionally, I've made both the slanted angle of the stripes and the direction to be mirrored as well while its RTL

components/progress-bar/ProgressBar.tsx
andrewgos: Good point. Initially i thought that min/max (like how React Bootstrap does it) doesn't add much value when it ended up being simply a bar again.

But from the accessibility POV, it makes sense to use real values.

Added min-max prop and adjust the unit test for that 👍

components/progress-bar/ProgressBar.tsx
andrewgos: Nice catch. Updated the warning to be no longer for dev only.

andrewgos: UPDATE:
I misread your comment earlier, thinking that it's about the 'only in development' issue.

But checking further, we had all other components to be checked in a11y on a dev mode only. For consistency, I've made it back to dev mode only, while applying the fix to no longer be limited to certain `labelVariant` only.

components/progress-bar/ProgressBar.tsx
Chocolate-lightning: I think that is a fair pov on my comment. I do however think if we were to leave it to the consumer, maybe the count description in SB and code comments can maybe detail our rational for not implementing a computed value / not providing a string formatter that does the calculations in the backend for them.

components/progress-bar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+const meta = {
+  title: 'Components/ProgressBar',
+  component: ProgressBar,
+  parameters: {
+    layout: 'padded',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    value: {
```
Chocolate-lightning, line 13: Whilst in practice this works well, in StoryBook when configuring custom min & max values, the value does not reflect the fill percentage correctly, i.e. with the set up of: value => 100, min => 20, max => 200 causes the fill to be 44.44% which is not quite right and the fill % can never reach 100% because of the SB prop limitations.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Great work so far, I think we are nearly there and there is only two tiny items to review before I merge this in.

components/progress-bar/ProgressBar.tsx
andrewgos: good point. Comments added into the component code and SB description on the min-max updated.

components/progress-bar/ProgressBar.stories.tsx
andrewgos: Thanks for pointing that out. Fixed the storybook to no longer using slider, more reflective of the freedom of what users can pass in but still normalised on the output

Chocolate-lightning (approved):
Thanks for taking my feedback on board. I think we are ready to go here.

PR #253, opened 2026-06-12
feat(Components)!: Entrypoint for the component-only styling assets

andrewgos:
## Description

This PR adds granular component styling entrypoints while keeping the existing full-build CSS behavior unchanged.

- Introduces component-only CSS and SCSS entrypoints, giving consumers finer control over which styles/assets they load (free from Bootstrap)
- Introduces per-component subpath imports, enabling consumers to import only what they use for better performance
- Preserving the existing`/dist/index.css` as the full-build CSS (Bootstrap + tokens + components) contract to load assets directly by URL
 
**BREAKING CHANGE:**
- Moves the published `/tokens` to `/dist/tokens`, ensuring all published artifacts are sourced from a single, consistent distribution boundary.

### How
A custom Vite generateBundle plugin emits styling assets during build.

The plugin:
- Emits `/dist/components/index.css` as an import manifest of components-only CSS files
- Emits `/dist/components/index.scss` as the same manifest for SCSS consumption
- Emits `/dist/components/<component>/index.css` files for each referenced component
- Emits token CSS/SCSS files into dist/tokens

Build behavior for the full bundle is intentionally unchanged:
- `cssCodeSplit` remains `false`
- The main CSS output remains one full inlined `/dist/index.css` bundle

### Package exports
New exports:
- `@moodlehq/design-system/components/css` → `/dist/components/index.css` (components-only)
- `@moodlehq/design-system/components/scss` → `/dist/components/index.scss` (components-only)

Updated token exports:
- `@moodlehq/design-system/tokens/css` → `/dist/tokens/index.css` (tokens-only)
- `@moodlehq/design-system/tokens/scss` → `/dist/tokens/_index.scss` (tokens-only)
- `@moodlehq/design-system/tokens/scss/legacy` → `/dist/tokens/_index.legacy.scss` (tokens-only)

Unchanged:
- `@moodlehq/design-system/css` → `/dist/index.css` (full bundle = Bootstrap + tokens + components)
- Component JS subpath exports remain JS/TS entrypoints only without CSS

### Why this approach
- Keeps the existing full-build contract safe for direct URL loading
- Avoids Bootstrap/token dependency-order issues that can happen with CSS splitting
- Adds component-only styling options without breaking existing consumers
- Keeps published package contents consistent under dist for clearer distribution boundaries

## Related Jira or GitHub Issue

Main task to enable component styling in LMS: https://moodle.atlassian.net/browse/MDS-595

Related tasks:
[...]

PR #256, opened 2026-06-17
feat(Link): Add new link component

Chocolate-lightning:
## Description

Adds the `Link` component to the Moodle Design System.

The component renders a native `<a>` anchor element styled with `--mds-*` tokens and Bootstrap-compatible class hooks. It supports two visual variants (`primary` and `secondary`), optional prefix/suffix icons, a disabled state, RTL layout, and keyboard focus.

Key implementation decisions:

- **No children** — label text is always passed as a named `label` prop, keeping the API explicit and i18n-safe.
- **Icon guard** — `startIcon` and `endIcon` accept only intrinsic `<i>` or `<svg>` elements. Invalid elements are dropped at runtime with a `console.error` in development, preventing arbitrary content from leaking into the icon slot.
- **Single icon rule** — if both `startIcon` and `endIcon` are provided the component renders only `startIcon` and emits a `console.warn`.
- **Disabled state** — anchors have no native disabled attribute. The component removes `href`, sets `aria-disabled="true"`, sets `tabIndex={-1}`, and blocks `onClick` via `event.preventDefault()` / `event.stopPropagation()`.
- **Focus ring** — uses a bottom-only `border-bottom` underline approach to match the Figma spec, which differs from the standard `outline` ring pattern used on other interactive components. `outline: none` is set on `:focus`, and the underline is applied on `:focus-visible`.
- **Icon animation** — prefix/suffix icons translate toward the label text on hover using `transform: translateX()` with logical-property-aware RTL reversals, so the link's layout width stays constant.
- **`forwardRef`** — the anchor element is forwarded via `forwardRef` to support form libraries and programmatic focus.

https://moodle.atlassian.net/browse/MDS-551

## Type of Change

- [x] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

## Accessibility notes

- The `<a>` element carries an implicit `role="link"` and is always accessible by keyboard.
- Disabled links retain `role="link"` and expose `aria-disabled="true"` so screen readers announce the element and its unavailable state rather than skipping it entirely.
- `tabIndex={-1}` removes disabled links from the sequential focus order while keeping them reachable programmatically.
- Icons are wrapped in `aria-hidden="true"` `<span>` containers and never carry meaningful text.
- The `label` prop is required and must be a caller-supplied translated string; raw string literals are not used inside the component.
- Focus ring uses `border-bottom` rather than `outline` per the Figma spec. This is a documented deviation from the standard MDS focus ring pattern — it applies only to the Link component and only to the `:focus-visible` pseudo-class.

## Files changed

| File | Notes |
|---|---|
| `components/link/Link.tsx` | Component implementation |
| `components/link/link.css` | Component-scoped styles using `--mds-*` tokens |
| `components/link/Link.stories.tsx` | Storybook stories: Default, Variants, WithIcons, Disabled, RightToLeft, StateMatrix |
| `components/link/Link.test.tsx` | Vitest unit tests |
| `components/link/Link.figma.tsx` | Figma Code Connect mapping (not shipped in package build) |
| `components/link/index.tsx` | Local barrel re-export |
| `components/index.tsx` | `Link` and `LinkProps` added to the public barrel |
| `components/index.css` | `@import './link/link.css'` added |

## Additional Notes

The focus ring implementation for this component deviates from the standard `outline` + `outline-offset` pattern documented in the component instructions. The Figma design specifies a bottom-only underline focus indicator (2 px `border-bottom`, 1 px gap). This is intentional and matches the visual design for links specifically, where an underline ring is semantically more appropriate than an enclosing box ring. The deviation is scoped to `Link` only and does not affect any other component.

### DIFF
--- a/components/link/Link.stories.tsx
+++ b/components/link/Link.stories.tsx
@@ -1,0 +1,292 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { expect, fireEvent, fn } from 'storybook/test';
+import { Link } from './Link';
+
+const iconMapping = {
+  None: undefined,
+  'Arrow Left': <i className="fa-solid fa-arrow-left" aria-hidden="true" />,
+  'Arrow Right': <i className="fa-solid fa-arrow-right" aria-hidden="true" />,
+};
+
+const meta = {
+  title: 'Components/Link',
+  component: Link,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    label: {
+      description: 'Visible link label.',
+      table: {
+        defaultValue: { summary: '' },
+      },
+    },
+    href: {
+      control: { type: 'text' },
+      description: 'Native anchor href.',
+    },
+    variant: {
+      control: { type: 'select' },
+      options: ['primary', 'secondary'],
+      description: 'Link visual variant.',
+      table: {
+        type: { summary: 'primary | secondary' },
+        defaultValue: { summary: 'primary' },
+      },
+    },
+    startIcon: {
+      description:
+        'Icon rendered before the label. Accepts only intrinsic `<i>` or `<svg>` elements.',
+      options: Object.keys(iconMapping),
+      mapping: iconMapping,
+      control: { type: 'select' },
+    },
+    endIcon: {
+      description:
+        'Icon rendered after the label. Accepts only intrinsic `<i>` or `<svg>` elements.',
+      options: Object.keys(iconMapping),
+      mapping: iconMapping,
+      control: { type: 'select' },
+    },
+    disabled: {
+      control: { type: 'boolean' },
+      description: 'Disables interaction and removes the href target.',
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    target: {
+      control: { type: 'select' },
+      options: ['_self', '_blank', '_parent', '_top'],
+      description:
+        'Browsing context for the link. When set to `_blank` the component automatically adds `rel="noopener noreferrer"` to prevent reverse tabnabbing.',
+      table: {
+        type: { summary: '_self | _blank | _parent | _top' },
+        defaultValue: { summary: '_self' },
+      },
+    },
+    rel: {
+      control: { type: 'text' },
+      description:
+        'Space-separated list of link relationship tokens (e.g. `noopener noreferrer`). When `target="_blank"` is set, `noopener` and `noreferrer` are always merged in automatically.',
+      table: {
+        type: { summary: 'string' },
+      },
+    },
+    onClick: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+  args: {
+    label: 'Link',
+    href: '#storybook-link',
+    variant: undefined,
+    disabled: false,
+  },
+} satisfies Meta<typeof Link>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    // Prevent the anchor from navigating during Storybook/Playwright tests.
+    // Without preventDefault the browser connection closes mid-play and the
+    // test runner reports a false "page closed unexpectedly" error.
+    onClick: fn((e: React.MouseEvent<HTMLAnchorElement>) => e.preventDefault()),
+  },
+  play: async ({ canvas, args, userEvent }) => {
+    const link = canvas.getByRole('link', { name: args.label });
+    await userEvent.click(link);
+    await expect(args.onClick).toHaveBeenCalledTimes(1);
+  },
+};
+
+export const Variants: Story = {
+  parameters: {
+    controls: { disable: true },
+    docs: { canvas: { sourceState: 'none' } },
+  },
+  render: () => (
+    <div style={{ display: 'flex', gap: 'var(--mds-spacing-xl)' }}>
+      <Link label="Primary" href="#storybook-link" />
+      <Link label="Secondary" href="#storybook-link" variant="secondary" />
+    </div>
+  ),
+};
+
+export const WithIcons: Story = {
+  parameters: {
+    controls: { disable: true },
+    docs: { canvas: { sourceState: 'none' } },
+  },
+  render: () => (
+    <div style={{ display: 'flex', gap: 'var(--mds-spacing-xl)' }}>
+      <Link
+        label="Start icon"
+        href="#storybook-link"
+        startIcon={iconMapping['Arrow Left']}
+      />
+      <Link
+        label="End icon"
+        href="#storybook-link"
+        endIcon={iconMapping['Arrow Right']}
+      />
+    </div>
+  ),
+};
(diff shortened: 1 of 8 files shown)


components/link/index.tsx
```diff
@@ -0,0 +1,3 @@
+// CSS is aggregated via components/index.css and imported by the package root entry (index.ts).
```
andrewgos: This comment will be no longer be true in https://github.com/moodlehq/design-system/pull/253 and can be removed

components/link/Link.tsx
```diff
@@ -0,0 +1,121 @@
...
+      ? (variant as LinkVariant)
+      : 'primary';
+  const resolvedStartIcon = isIconElement(startIcon, 'startIcon')
+    ? startIcon
+    : null;
+  const resolvedEndIcon = isIconElement(endIcon, 'endIcon') ? endIcon : null;
+
+  if (import.meta.env.DEV) {
+    const hasAccessibleName =
+      Boolean(label) ||
```
andrewgos: for accessibility checking, the `Boolean()` check might be misleading when it becomes truthy for whitespaces " " while it's being inaccessible.

perhaps doing a check on string length is safer `label.trim().length > 0`

components/link/Link.tsx
```diff
@@ -0,0 +1,121 @@
...
+  if (className) {
+    classes.push(className);
+  }
+
+  return (
+    <a
+      ref={ref}
+      className={classes.join(' ')}
+      href={disabled ? undefined : href}
+      aria-disabled={disabled || undefined}
```
andrewgos, line 117: as this component has a custom 'disabled' functionality, we may need to be stricter with the order of props being spread.

If the user passed on `aria-disabled` as `false` in the props, it'll be overriding this resolved disabled value, leaving us with enabled in the ARIA but shows a disabled behaviour with the JS/CSS

components/link/Link.tsx
```diff
@@ -0,0 +1,121 @@
...
+      tabIndex={disabled ? -1 : tabIndex}
+      role={disabled ? (role ?? 'link') : role}
+      onClick={handleClick}
+      {...props}
```
andrewgos: We may only be using the native `target` prop, but we might want to consider the safety guard whenever there is a `target="_blank"` being used, that is to use the `rel` feature with `noopener` and `noreferrer` to avoid the new tabs from attacking the original tab (reverse tabnabbing)

components/link/Link.stories.tsx
```diff
@@ -0,0 +1,274 @@
...
+] as const;
+
+export const StateMatrix: Story = {
+  parameters: {
+    controls: { disable: true },
+    docs: {
+      canvas: { sourceState: 'none' },
+      description: {
+        story:
+          'State matrix for visual regression review across variant and icon usage combinations. Hover, pressed, and focus-visible cells are driven by the Storybook pseudo-states addon.',
```
andrewgos, line 236: i dont see any pseudo states classes in the preview.css yet?

components/link/Link.stories.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+const linkMatrixStates = [
+  { key: 'default', label: 'Default' },
+  { key: 'hover', label: 'Hover' },
+  { key: 'pressed', label: 'Pressed' },
+  { key: 'focus-visible', label: 'Focus visible' },
+  { key: 'disabled', label: 'Disabled' },
+] as const;
+
+export const StateMatrix: Story = {
```
andrewgos, line 229: What does the **pressed** state is actually trying to show?

With the same color as the hover, shouldnt the pressed look the same with the closer icon and underline too?

components/link/link.css
```diff
@@ -0,0 +1,122 @@
...
+.mds-link__icon i,
+.mds-link__icon svg {
+  inline-size: var(--mds-icons-xs);
+  block-size: var(--mds-icons-xs);
+  font-size: var(--mds-icons-xs);
+  line-height: 1;
+  flex-shrink: 0;
+}
+
+.mds-link__icon {
```
andrewgos, line 113: I couldn't tell whats wrong, but to me it seems to shift vertically when the icon is added?

https://github.com/user-attachments/assets/9d1fe8ad-d890-4a12-8c8f-4f895f57ee83

Also, Is this 8px border intended? I couldnt see it in Figma
Image

andrewgos (changes requested):
Nice work on tackling this one so quickly!

It's in a very good shape, there's only a few minor items that I'd like to pointed out in regards to security and questions with the styling.

components/link/Link.stories.tsx
```diff
@@ -0,0 +1,274 @@
...
+  'Arrow Right': <i className="fa-solid fa-arrow-right" aria-hidden="true" />,
+};
+
+const meta = {
+  title: 'Components/Link',
+  component: Link,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'beta'],
```
andrewgos: one more thing: `beta` to `stable` before release

components/link/index.tsx
Chocolate-lightning: Good catch!

components/link/Link.tsx
Chocolate-lightning: Great suggestion, rolling that in in.

components/link/Link.tsx
Chocolate-lightning: {...props} moved before computed props, plus a new test confirming a consumer-passed aria-disabled={false} cannot override the component's accessibility logic.

components/link/Link.tsx
Chocolate-lightning: Added resolvedRel that auto-injects noopener noreferrer when target="_blank", with 3 tests covering: no rel, existing rel (dedup), and non-blank target.

components/link/Link.stories.tsx
Chocolate-lightning: They are not required here and it is more for complex components.

components/link/Link.stories.tsx
Chocolate-lightning: Yea I find it a bit odd as well but that is more of a design point that we can discuss with them as they do the UI review.

Chocolate-lightning: Pressed state is working as expected as per Boris' design and request.

components/link/link.css
andrewgos: As discussed, this seems to be an intermittent issue caused by the Storybook wrapper, not the component itself. So we can pass this one and monitor it, we can re-investigate if this becomes a major issue.

Chocolate-lightning: Padding is coming from StoryBook itself for the zooming functionality. Vertical shifting can not be replicated across different browsers or builds. If we spot it out in the wild we can address it as a small patch down the line.

andrewgos (approved):
As we are passing on the Storybook issue, I believe we can proceed to merge ✅ We can always revisit/reinvestigate as a follow-up patch if the Storybook wrapper is causing a major issue.

PR #259, opened 2026-06-19
feat(FavouriteButton): Add new FavouriteButton component

andrewgos:
## Description

Adds the new `FavouriteButton` component for marking an item as favourited or removing it from favourites.

The component is an icon-only toggle button using the documented star pattern: an outlined star when unselected and a filled star when selected. It exposes a controlled `selected` prop, requires an `aria-label` from the caller, and sets `aria-pressed` so assistive technologies can announce the toggle state.

### Key implementation decisions

- Implemented `FavouriteButton` as a native `<button>` with `forwardRef`, so consumers can access the underlying button element.
- Required `aria-label` as a prop because the component is icon-only and has no visible text label.
- Used `aria-pressed={selected}` to communicate the toggle state instead of relying on the filled/outlined star alone.
- Omitted the native `type` prop from the public props interface and hardcoded `type="button"` to avoid accidental form submission.
- Used Font Awesome star classes already available in the design system: `fa-regular fa-star` for unselected and `fa-solid fa-star` for selected.
- Marked the icon as decorative with `aria-hidden="true"` because the accessible name comes from the button label.
- Added stable `mds-favourite-button` and `mds-favourite-button--active` class hooks, with styling driven by existing `--mds-*` tokens.
- Added Storybook stories for default, selected, and disabled states, including interaction/a11y assertions.
- Added unit tests for class names, accessibility attributes, icon rendering, disabled behavior, prop forwarding, click handling, default button type, and ref forwarding.
- Added Code Connect mappings for Figma variants across selected/unselected and documented interaction states.
- Exported the component and props from both the component folder and the package component barrel, and imported the stylesheet into `components/index.css`.

https://moodle.atlassian.net/browse/MDS-552

## Screenshots/Previews

Screenshot 2026-06-19 at 11 07 24 pm

- [ ] `ComponentName.figma.tsx` Code Connect file created and published

### TypeScript and props

### Internationalisation

- [ ] RTL story added if the component has directional layout

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes

This is a new component addition only. No existing component props, exports, tokens, or package paths were removed or renamed.


components/favourite-button/FavouriteButton.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+  },
+};
+
+const showcaseStyle = {
+  display: 'flex',
+  alignItems: 'center',
+  gap: 'var(--mds-spacing-sm)',
+};
+
+export const Default = {} satisfies Story;
```
Chocolate-lightning: I can't toggle the selected state when pressing on the fav button in my local SB instance, We should ensure the component can be fully interacted with in SB.

components/favourite-button/favourite-button.css
```diff
@@ -0,0 +1,72 @@
...
+  border-radius: var(--mds-border-radius-pill);
+  background-color: var(--mds-bg-surface-default);
+  color: var(--mds-text-default);
+  cursor: pointer;
+  transition:
+    background-color 0.15s ease-in-out,
+    color 0.15s ease-in-out;
+}
+
+.mds-favourite-button .fa-regular,
```
Chocolate-lightning: I am seeing visual inconsistencies between implementation and design, In the designs it seems the icon is clamped into being 12px x 12px but in implementation I am seeing 15px x 12px. Has the spec changed since I last had a look or is there something that needs to be done here?

components/favourite-button/FavouriteButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      ref={ref}
+      className={classes}
+      type="button"
+      aria-label={ariaLabel}
+      aria-pressed={selected}
+      {...props}
+    >
+      {/* Star icon: filled when selected, outlined when unselected */}
+      <i
+        className={selected ? 'fa-solid fa-star' : 'fa-regular fa-star'}
```
Chocolate-lightning: This implementation forces the usage of F/A. I suggest that we may need to do something similar to what we did with the close button. We may need to look at seeing how we can credit F/A if we use the icon in such a manner.

components/favourite-button/favourite-button.css
```diff
@@ -0,0 +1,72 @@
...
+.mds-favourite-button--selected:active:not(:disabled) {
+  background-color: var(--mds-bg-surface-strong);
+  color: var(--mds-text-link-primary-default);
+}
+
+/* Disabled states */
+.mds-favourite-button:disabled {
+  background-color: var(--mds-bg-surface-default);
+  color: var(--mds-text-muted);
+  cursor: not-allowed;
```
Chocolate-lightning, line 69: Was this cursor definition defined somewhere in the docs? I note that the button component has the pointer cursor for the hover state which is not defined by use. This is emulated earlier in this component styling but the disabled state does not use the not-allowed in the button component.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Overall a nice and simple button approach with little that needs amending before merge.

Cheers,

components/favourite-button/FavouriteButton.stories.tsx
andrewgos: amazing feedback. updated the SB to be interactive.

meanwhile, also removed 'Focus' story as that (and pseudo-state matrix) may not be adding much value

components/favourite-button/favourite-button.css
andrewgos: Nicely spotted, that also helped me find out that we shouldn't use FontAwesome as a dependency (like the close button), it's causing the misalignment if we aren't doing sizing on the SVG itself.

Changing the FA implementation to be a local SVG copy

components/favourite-button/FavouriteButton.tsx
andrewgos: agree. changing FA to SVGs

this was causing the 12x12 misalignment as well because the sizing was limited to font-size only when using FA

components/favourite-button/favourite-button.css
andrewgos: Yes it was as documented in the Zeroheight

https://design.moodle.com/98292f05f/p/932a22-favourite-button/t/f4e33b3ef5

components/favourite-button/assets/star-filled.svg
Chocolate-lightning, line 1: I think we should likely move these two svg files into a /assets location same as we have done for the activityicon component. If we were to move to a pattern like this, we should also update the closebutton svg location but this can be done at a later date.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

We are nearly there and I only have a couple of comments.

A new entry also needs to be added to .github/instructions/component-index.instructions.md

I am also a bit concerned about the uplift project work as looking at the MSE it uses a heart icon but we only ship a star icon here, I suppose it comes back to this originally being a toggle button rather than a small favouritebutton, not really an action item but wanted to note it.

Cheers,

components/favourite-button/assets/star-filled.svg
andrewgos: good point. I've moved it and renamed it to `assets/star-filled.svg` and `assets/star-outline.svg` for scalability if we have more icons for favourite button

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Thanks for taking my feedback onboard. I see just one item remaining before we can look at merging this in.

After testing the MDS package with the fav button included within the build inside Moodle Core LMS, I am seeing the same issue that we have with the close button not including the required asset. The assets are not included in the build and thus do not come across into LMS. 

Subset of the output from `npm pack`
Image

Hovering over the new favourite button component at the bottom of the page in an LMS demo page.
Image

Cheers,

Chocolate-lightning:
As discussed, we will accept that the icon svg is not loading correctly as it is similar to the issue we are hitting in the close button component and a more general fix may be required. I have raised https://github.com/moodlehq/design-system/issues/264 as a followup to address this issue but for now we can merge this in once the branch is rebased.

Cheers,

andrewgos:
Thanks for checking and pointing that out @Chocolate-lightning .

I can see this was an existing issue happening DS-wide as a published production package (which is happening to the close button as well), so I'd lean towards keeping this as a separate issue/ticket in the next sprint. And with that, we can have more thinking time on the fix for the close-button as well.

Chocolate-lightning (approved):
Approving this as all tests are green and feedback addressed. Don't forget to update the ZH site once we get the release published and update the MSE checklist.

PR #261, opened 2026-06-22
feat(NavPill): Add new NavPill component

andrewgos:
## Description

Adds the `NavPill` component — a compact navigation destination button used to identify and navigate between top-level sections of a UI. The pill renders a visible label and optionally an active-indicator dot when `selected`.

Key behaviours:

- `selected=true` marks the pill as the active navigation item, setting `aria-current="page"` and rendering an indicator dot
- A selected pill **cannot** be disabled — the `disabled` prop is silently ignored when `selected=true` to ensure the current location is always identifiable (dev-mode warning emitted)
- `aria-current` is intentionally applied after `{...props}`, and `aria-current` is excluded from `NavPillProps` to preserve active-state semantics.

https://moodle.atlassian.net/browse/MDS-555

## Screenshots/Previews

Screenshot 2026-06-22 at 1 56 23 pm

### TypeScript and props

- [ ] `...props` spread last on the host element
  > `aria-current` is intentionally applied after `{...props}`, and `aria-current` is excluded from `NavPillProps` to preserve active-state semantics.

### Internationalisation

### CSS and tokens

  > `.mds-nav-pill__indicator` intentionally uses hardcoded `width: 4px; height: 4px;` as per Figma

### Tests and stories

### Breaking changes

### AI context

- [ ] Instruction files updated if this component introduces new patterns or anti-patterns
  > The intentional `{...props}`-before-overrides pattern (for safety-critical attributes) is a new pattern not currently documented in the instruction files.

## Cross-system parity

## Additional Notes
- Property-based fuzz testing via `fast-check` (`fuzzComponent`) is included across 100 random prop combinations
- The `selected + disabled` interaction (selected always wins) is covered by dedicated unit tests and a dev-mode `console.warn`


components/nav-pill/NavPill.tsx
Chocolate-lightning, line 1: This component needs to be referenced in .github/instructions/component-index.instructions.md like the others.

components/nav-pill/NavPill.stories.tsx
Chocolate-lightning, line 1: This component may benefit from a pseudo-state matrix like other components that contain hover, pressed & focus states.

components/nav-pill/NavPill.figma.tsx
```diff
@@ -0,0 +1,48 @@
...
+figma.connect(NavPill, url, {
+  variant: { State: 'Disabled', Selected: 'No' },
+  props: {
+    label: label,
+  },
+  example: ({ label }) => <NavPill label={label} disabled />,
+});
+
+// Disabled selected
+figma.connect(NavPill, url, {
```
Chocolate-lightning, line 28: Looking at the Figma designs, I do not see this case. Was this something that was generated based on some loose documentation or reference in the component info or ZH docs?

components/nav-pill/NavPill.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+/** Default unselected pill — the resting state for inactive navigation items. */
+export const Default: Story = {
+  args: {
+    label: 'Label',
+    selected: false,
+  },
+};
+
+/** All states displayed together for a visual overview. */
+export const States: Story = {
```
Chocolate-lightning, line 94: This story & tests could be expanded to include a full nav group to test the tabbing functionality documented in https://design.moodle.com/98292f05f/p/981bd1-nav-pill/t/37bcbe9a42 or some comments and documentation should be added into the package or story about how this singular component should be wrapped in a parent element.

components/nav-pill/NavPill.tsx
```diff
@@ -0,0 +1,113 @@
...
+      rel,
+      tabIndex,
+      ...props
+    },
+    ref,
+  ) {
+    // A selected pill represents the active navigation destination and cannot
+    // be disabled — disabling it would leave users with no way to identify
+    // the current location. Silently ignore the disabled prop when selected.
+    if (import.meta.env.DEV && selected && disabled) {
```
Chocolate-lightning, line 49: We should have a unit test confirming this warning which will be inline with other components that do similar runtime validation.

components/nav-pill/NavPill.tsx
```diff
@@ -0,0 +1,113 @@
+import type { AnchorHTMLAttributes, MouseEvent } from 'react';
+import { forwardRef } from 'react';
+
+export interface NavPillProps extends Omit<
+  AnchorHTMLAttributes<HTMLAnchorElement>,
+  'href' | 'aria-current' | 'role'
```
Chocolate-lightning: NavPillProps omits href, aria-current, and role to prevent consumers from overriding safety-critical attributes. aria-disabled is also safety-critical (the component's disabled semantics depend on it) but it is not in the Omit list.

The component overrides aria-disabled post-spread, but the type surface is misleading — a consumer who reads the types will believe they can control aria-disabled directly.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Overall I think this is a solid approach. I have added my thoughts against certain files and I have a couple of other points that may need to be addressed but I'll hold off on them until my question in the Figma file is answered.

Cheers,

Chocolate-lightning:
Also, dont forget about the navpill token updates Boris published last week in https://github.com/moodlehq/design-system/pull/258

components/nav-pill/nav-pill.css
```diff
@@ -0,0 +1,96 @@
...
+
+.mds-nav-pill:focus-visible {
+  outline: var(--mds-stroke-weight-md) solid var(--mds-focus-default);
+  outline-offset: var(--mds-spacing-offset);
+  box-shadow: none;
+}
+
+/* ---- Disabled ---- */
+
+.mds-nav-pill:disabled,
```
Chocolate-lightning: This definition occurs a few times in the stylesheet, but `:disabled` is not a valid pseudo-state for links and `[aria-disabled='true'` is doing the heavy lifting. Maybe we can consider removing these :disabled definitions that are inert

components/nav-pill/nav-pill.css
```diff
@@ -0,0 +1,96 @@
+/* Nav pill — compact navigation destination indicator */
+
+.mds-nav-pill {
+  /* Reset button defaults */
```
Chocolate-lightning: We are using links in this component rather than a button base. This comment should probably better reflect the actual state.

components/nav-pill/nav-pill.css
```diff
@@ -0,0 +1,96 @@
...
+  gap: var(--mds-spacing-xxs);
+  padding: var(--mds-spacing-xs) var(--mds-spacing-sm);
+  border: none;
+  background-color: transparent;
+  cursor: pointer;
+
+  border-radius: var(--mds-border-radius-md);
+
+  /* Label colour: unselected default */
+  color: var(--mds-text-subtle);
```
Chocolate-lightning, line 17: We may have to check the specificity of this definition as it has (0,1,0) but if someone declares `a:visited { color: … }` like bootstrap does, the rule then has a specificity of (0,1,1) and will win, leaking visited-link colour onto pills the user has previously navigated to.

Is this desired or just a potential issue? May be worth consulting the designers here as well.

Chocolate-lightning (changes requested):
Just a few more points after discussing the potential of nav-pill dropdowns.

components/nav-pill/nav-pill.css
andrewgos: cheers. this must be the leftovers when i migrated from <button> to <a>. removing it

components/nav-pill/nav-pill.css
andrewgos: updating the comment 👍

components/nav-pill/nav-pill.css
andrewgos: confirmed with Edu, we are not doing visited state for the NavPill. I'm adding restriction onto the css to ensure visited state remains the same

components/nav-pill/NavPill.tsx
andrewgos: updating the instruction file 👍

components/nav-pill/NavPill.stories.tsx
andrewgos: I was thinking this NavPill component is too simple to have pseudo-state matrix, but I don't mind having it either 👍  Adding one in

andrewgos: Thinking further... i decided the matrix wont be much useful, because 'selected' state do not have hover/pressed/disabled state.

Let me know if you can think of another reason if having matrix is worth here

components/nav-pill/NavPill.figma.tsx
andrewgos: good catch thanks!! yes it does not exist because selected will override the disabled state. removing this

andrewgos: tidied up the selected state as well, visual and code representation of 'selected' state will be the same for hover/selected/disabled

components/nav-pill/NavPill.stories.tsx
andrewgos: Good point, but I would argue that those demonstrations should be documented in the storybook of the parent component that wraps it (if we came up with a NavGroup component, for example) instead of the individual atomic component. Because I see it more as an implementation needed to be done on the parent/container side.

Added unit test for the behavior of the group integration tho, checking for its onFocus and tabIndex behaviour.

wdyt?

components/nav-pill/NavPill.tsx
andrewgos: added unit test for this 👍

components/nav-pill/NavPill.tsx
andrewgos: added `aria-disabled` to the omit list 👍

components/nav-pill/NavPill.stories.tsx
Chocolate-lightning: It would only really benefit to have a focus state somewhere in the stories but considering the outline implementation is in line with others, I think we can omit it for a small component.

components/nav-pill/NavPill.stories.tsx
Chocolate-lightning: Sounds like a good compromise and the tests look good.

Chocolate-lightning (approved):
All action items are done and everything looks good both locally and in CI so I will accept this in now.

Great work!

PR #263, opened 2026-06-22
feat(Choicebox): Introduce Choicebox as a detailed radio input type

Chocolate-lightning:
## Description

This PR introduces the new `Choicebox` component to the Moodle Design System.

The implementation follows the component checklist workflow and aligns naming/behaviour across code, Storybook, and Figma. The main goal is to provide a reusable, accessible choicebox primitive that can be consumed consistently in Moodle interfaces while preserving token-driven styling and predictable keyboard/screen-reader behaviour.

### General approach

1. Design-first parity

- Use the Figma source as the behaviour and visual reference baseline: https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=10232-4215&m=dev

2. API and rendering model

- Ensure invalid/unsupported prop values gracefully fall back to documented defaults.

3. Styling and token compliance

- Use existing `--mds-*` tokens and preserve the project guardrail of no fallback literals where matching tokens exist.
- Keep variants composable via stable class hooks (`mds-*`) and predictable CSS structure.

4. Accessibility and interaction quality

- Validate focus visibility and keyboard navigation behaviour in stories/tests.

5. Verification and rollout readiness

- Cover visual/interactive states via Storybook stories and test tags.
- Use the MDS build/checklist guidance as a readiness gate before merge.

- Jira: https://moodle.atlassian.net/browse/MDS-574
- Confluence checklist/reference: https://moodle.atlassian.net/wiki/spaces/MDS/pages/4007657524/Choicebox+MSE+Build+Checklist
- Figma design reference: https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=10232-4215&m=dev

### TypeScript and props

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

### DIFF
--- a/components/choicebox/Choicebox.figma.tsx
+++ b/components/choicebox/Choicebox.figma.tsx
@@ -1,0 +1,184 @@
+import figma from '@figma/code-connect';
+import { Choicebox } from './Choicebox';
+
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=10232-4215';
+
+const baseProps = {
+  label: figma.string('Label'),
+  supportingText: figma.string('Support text'),
+};
+
+// -----------------------------------------------------------------------
+// Unselected — no supporting text
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { Selected: 'No', 'Supporting text': false },
+  props: {
+    label: figma.string('Label'),
+  },
+  example: ({ label }) => <Choicebox label={label} />,
+});
+
+// -----------------------------------------------------------------------
+// Unselected — with supporting text
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { Selected: 'No', 'Supporting text': true },
+  props: baseProps,
+  example: ({ label, supportingText }) => (
+    <Choicebox label={label} supportingText={supportingText} />
+  ),
+});
+
+// -----------------------------------------------------------------------
+// Selected — no supporting text
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { Selected: 'Yes', 'Supporting text': false },
+  props: {
+    label: figma.string('Label'),
+  },
+  example: ({ label }) => <Choicebox label={label} checked />,
+});
+
+// -----------------------------------------------------------------------
+// Selected — with supporting text
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { Selected: 'Yes', 'Supporting text': true },
+  props: baseProps,
+  example: ({ label, supportingText }) => (
+    <Choicebox label={label} supportingText={supportingText} checked />
+  ),
+});
+
+// -----------------------------------------------------------------------
+// Disabled (unselected)
+// Disabled state is a separate variant split so the disabled prop
+// appears explicitly in the Dev Mode snippet without conditional logic.
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { State: 'Disabled', Selected: 'No' },
+  props: {
+    label: figma.string('Label'),
+    supportingText: figma.string('Support text'),
+  },
+  example: ({ label, supportingText }) => (
+    <Choicebox label={label} supportingText={supportingText} disabled />
+  ),
+});
+
+// -----------------------------------------------------------------------
+// Disabled (selected)
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { State: 'Disabled', Selected: 'Yes' },
+  props: {
+    label: figma.string('Label'),
+    supportingText: figma.string('Support text'),
+  },
+  example: ({ label, supportingText }) => (
+    <Choicebox label={label} supportingText={supportingText} checked disabled />
+  ),
+});
+
+// -----------------------------------------------------------------------
+// Unselected — icon, no supporting text
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { Selected: 'No', 'Supporting text': false, Icon: true },
+  props: {
+    label: figma.string('Label'),
+  },
+  example: ({ label }) => (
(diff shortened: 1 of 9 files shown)


components/choicebox/Choicebox.stories.tsx
```diff
@@ -0,0 +1,305 @@
...
+      control: { type: 'text' },
+      description:
+        'ID for the input element. If omitted, a unique ID is generated automatically.',
+      table: {
+        type: { summary: 'auto-generated | string' },
+        defaultValue: { summary: 'auto-generated' },
+      },
+    },
+  },
+  args: {
```
andrewgos, line 97: we can add
```
checked: false,
disabled: false
```

for a better SB experience on a default state of the controls

components/choicebox/Choicebox.stories.tsx
```diff
@@ -0,0 +1,305 @@
...
+        type: { summary: 'string' },
+        defaultValue: { summary: 'undefined' },
+      },
+    },
+    icon: {
+      description:
+        'Optional icon rendered before the label group. Must be an `<i>` or `<svg>` element.',
+      control: false,
+      table: {
+        type: { summary: 'ReactElement<"i" | "svg">' },
```
andrewgos, line 49: i understand this can be anything, but perhaps just for SB demonstration purposes we provide some sample icons to play around with? Check out the Button.stories.tsx for reference

components/choicebox/Choicebox.stories.tsx
```diff
@@ -0,0 +1,305 @@
...
+        defaultValue: { summary: 'undefined' },
+      },
+    },
+    checked: {
```
andrewgos, line 53: looks like the current SB doesnt let the interactivity with the controls. Was expecting that I would see `checked` in the code preview when i checked it

With the ability to uncheck, i suppose thats not allowed unless we click another choicebox right?

components/index.css
```diff
@@ -1,4 +1,5 @@
 @import './activity-icon/activity-icon.css';
+@import './choicebox/choicebox.css';
```
andrewgos: lets sort this alphabetically? :) should be under `checkbox` before `close-button`

components/index.tsx
```diff
@@ -1,6 +1,9 @@
 export { ActivityIcon } from './activity-icon';
 export type { ActivityIconProps } from './activity-icon';
 
+export { Choicebox } from './choicebox';
```
andrewgos: lets sort this alphabetically? :) should be under `checkbox` before `close-button`

components/choicebox/choicebox.css
```diff
@@ -0,0 +1,366 @@
...
+  background-color: var(--mds-bg-surface-strong);
+}
+
+/* Checked: filled circle with inline SVG checkmark via background-image.
+   The SVG path matches the circle-check glyph visible in the Figma design. */
+.mds-choicebox-input:checked + .mds-choicebox .mds-choicebox-indicator::after {
+  border: none;
+  background-color: var(--mds-bg-interactive-primary-default);
+  /* Inverted checkmark on filled primary background. */
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'%3E%3Cpath fill='%23ffffff' d='M6.5 11.5L2.5 7.5l1.06-1.06L6.5 9.38l5.94-5.94L13.5 4.5z'/%3E%3C/svg%3E");
```
andrewgos: the checkbox svg seems to be off from figma

Image

Image

Can you please re-export from figma to match it?

components/choicebox/Choicebox.figma.tsx
```diff
@@ -0,0 +1,184 @@
...
+const baseProps = {
+  label: figma.string('Label'),
+  supportingText: figma.string('Support text'),
+};
+
+// -----------------------------------------------------------------------
+// Unselected — no supporting text
+// -----------------------------------------------------------------------
+figma.connect(Choicebox, url, {
+  variant: { Value: 'Unselected', 'Supporting text': false },
```
andrewgos: i wonder why Figma does `Value: Selected/Unselected`, might be worth asking the designer of not why doing it like the other components `Selected: true/false` ?

andrewgos (changes requested):
Nice work! Here are some items that I can spot myself manually, I'm restraining myself from running AI checks as I'm approaching my credits limit..

components/choicebox/Choicebox.stories.tsx
Chocolate-lightning: Great suggestion!

components/choicebox/Choicebox.stories.tsx
Chocolate-lightning: Great suggestion!

components/choicebox/Choicebox.stories.tsx
Chocolate-lightning: Yea to interactively test out the state changes it has to be done when there is another element instance to swap to just like a radio input. This functionality is covered under the "Group" story.

components/choicebox/Choicebox.figma.tsx
edumoodle: Good call! we can do it, for sure. Proceed to update the component in component in Figma.

edumoodle: ✅  Done

components/choicebox/choicebox.css
Chocolate-lightning: What a keen eye you have, Updated to match the token svg definition.

components/choicebox/Choicebox.figma.tsx
barbararamiro: I agree with Andrew. We looked into something similar to this for the favourite button, and there are benefits in using the same approach here.

andrewgos:
Thanks for taking in the suggestions! I'm fine with this getting merged 👍

andrewgos (approved):

PR #268, opened 2026-06-26
feat(Avatar): Introduce the Avatar component

Chocolate-lightning:
## Description

Adds the `Avatar` component to the Moodle Design System.

The component displays a circular user avatar in six sizes (`xs`, `sm`, `md`, `lg`, `xl`, `xxl`). It uses a layered rendering strategy: initials are always present as the base layer, and a photo image is absolutely positioned on top. If the image fails to load, the component automatically falls back to the initials layer without a repaint — no consumer-side error handling required.

Key implementation decisions:

- **Image + initials are a unified model** — no `type` prop. The consumer sets `imageSrc` when a photo is available; the fallback is automatic. This diverges from the Figma component which models `Image` / `Initials` as separate variants; the design team should update the Figma component accordingly (see open question#1 below).
- **`xs` and `sm` are image-only sizes** per the design spec. Initials are suppressed at these sizes via `display: none` in CSS (the DOM node is kept so the fallback mount point is always present). A `console.warn` fires in `DEV` if `imageSrc` is absent at these sizes.
- **Accessible name follows the `` pattern** — consistent with MUI, Primer, React Spectrum, and Atlassian. `alt` is passed to the inner ``; the initials span is `aria-hidden` when an image is showing. No `role="img"` on the root span.
- **`forwardRef`** is implemented so consumers can pass refs for tooltip/popover anchoring without a wrapper element.
- **SVG silhouette fallback** (`assets/image-fallback.svg`) renders automatically whenever there is no valid image to display and no usable initials — specifically: (a) `initials=""` at any size, or (b) `xs`/`sm` at any time (initials are image-only sizes per spec). The silhouette is decorative (`alt=""`, `aria-hidden`) and sits in the same absolutely-positioned `mds-avatar__image` layer as a real photo.

### Files changed

| File | Change |
|---|---|
| `components/avatar/Avatar.tsx` | New component. `AvatarSize` union type exported. `AvatarProps` typed correctly. `forwardRef`, `displayName`, DEV guards, `alt` prop, initials fallback, SVG silhouette fallback. |
| `components/avatar/avatar.css` | New component styles. Six size variants via `--mds-avatars-*` tokens. `display: none` suppression for initials at `xs`/`sm`. `line-height: 1` exception documented inline. |
| `components/avatar/index.tsx` | Barrel: `export * from './Avatar'` |
| `components/avatar/Avatar.stories.tsx` | Stories: `Default`, `InitialsFallback`, `SilhouetteFallback`, `AllSizes`. Image row covers all six sizes; initials row covers `md–xxl` only. Each story has a `play` function with interaction assertions. |
| `components/avatar/Avatar.test.tsx` | Unit tests: hook class, size modifier classes, invalid-size fallback, DEV warnings, initials rendering, image rendering, `imgFailed` fallback, SVG silhouette fallback, `forwardRef`. |
| `components/avatar/Avatar.figma.tsx` | Figma Code Connect. Two variants (`Image`, `Initials`). `initials` bound via `figma.string('Initials')` — no hardcoded literal. |
| `components/index.tsx` | Re-exports `Avatar`, `AvatarProps`, `AvatarSize`. |
| `index.ts` | Top-level barrel updated (auto-discovered by Vite build — no manual change needed if the folder is present). |

---

<https://moodle.atlassian.net/browse/MDS-553>

<https://github.com/moodlehq/design-system/pull/266>

---

## Type of Change

- [x] Feature

---

## Screenshots/Previews

Storybook stories to review:

- `Components/Avatar` — Default (image)
- `Components/Avatar` — InitialsFallback (initials-only)
- `Components/Avatar` — SilhouetteFallback (empty initials → SVG placeholder)
- `Components/Avatar` — AllSizes (image row: all 6; initials row: md–xxl)

---

## Checklist

---

## Accessibility

- `alt` prop is passed to the inner `` element. When `imageSrc` is set, callers are expected to pass the user's full name as `alt` (e.g. `alt="Jane Doe"`). Omit or leave empty only when the avatar is purely decorative — this is enforced by a DEV warning.
- When the image is showing, the initials span is `aria-hidden={true}`. When acting as the fallback, it is exposed as plain text so screen readers can announce the initials.
- The root `<span>` carries no explicit ARIA role. Screen readers encounter the `` directly, which is the standard cross-library pattern (MUI, Primer, Atlassian).
- NVDA image-navigation mode (`I` key in browse mode) will correctly land on the `` with the provided `alt` text.
- ZeroHeight guideline met: *"Include the user's full name in the associated label attribute, not just initials."*

---

## Open Questions

These items were surfaced during peer review and are not blocking merge, but should be resolved or assigned before the component reaches a v1 stable tag.

**Q1 — Figma component model alignment**
The Figma component models two separate `type` variants (`Image` / `Initials`). The React implementation has no `type` prop; the fallback is automatic. The design team should update the Figma component to reflect the unified model. The Code Connect file covers both variants so Dev Mode snippets are accurate, but the Figma component's own variant structure still implies they are mutually exclusive choices rather than a single component with an optional image.

---

### Token usage
All size dimensions use the `--mds-avatars-*` token set (`--mds-avatars-xs` through `--mds-avatars-xxl`). Background, text colour, border-radius, font-family, and letter-spacing all use `--mds-*` tokens. One deliberate exception: `line-height: 1` on the initials span. All existing `--mds-line-height-*` tokens are absolute rem values designed for prose; using any of them would add a fixed vertical gutter that misaligns the glyph inside the fixed-size flex container. The exception is documented inline in `avatar.css`. A `--mds-line-height-none: 1` primitive token should be requested from the design team if this pattern recurs.

### Bootstrap 5 compatibility
No class-name or custom-property collisions. Bootstrap has no `.avatar` class; `--mds-*` is distinct from `--bs-*`. The component's CSS rules carry no `!important`, so they coexist cleanly with Bootstrap resets. One advisory: Bootstrap's `.rounded-circle` utility carries `!important border-radius: 50%`. If a developer mistakenly applies it to an avatar element it will override `--mds-border-radius-pill`. The visual difference at the largest sizes is negligible, but the behaviour is undocumented — worth a note in ZeroHeight usage docs.


components/avatar/avatar.css
```diff
@@ -0,0 +1,109 @@
+/*
+ * avatar.css
+ * Styles for the Avatar component.
+ *
+ * Sizing uses --mds-scale-* primitives because no semantic avatar-size tokens
```
andrewgos, line 5: i can see you got the tokens in this branch already (and it'll be coming with https://github.com/moodlehq/design-system/pull/270), so this comment is no longer valid and can be removed

components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const [imgFailed, setImgFailed] = useState(false);
+
+    const sizeValue: AvatarSize =
+      size && allowedSizes.includes(size) ? size : 'md';
+
+    // Show the image only when a src is provided and it has not errored.
+    const showImage = !!imageSrc && !imgFailed;
+
+    // Show the SVG silhouette when there is no valid image to display and
+    // either (a) no initials were supplied, or (b) the size is xs/sm where
```
andrewgos, line 67: I think `initials` probably need more work to get things more aligned:

- It mentions "no initials were supplied," while it is seen as a required field in Storybook since we didn't declare it as optional in the prop list. Perhaps consider making it optional, too?

Image

- However, I can see that people can enter an empty string, implying this "no initials were supplied", perhaps we can add validations for invalid initials? I'd imagine people can enter empty spaces, symbols, lower cases, etc. where it could break the look or break the "required" conditions

components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+
+    const classes = ['mds-avatar', `mds-avatar--${sizeValue}`];
+    if (className) classes.push(className);
+
+    return (
+      <span ref={ref} className={classes.join(' ')} {...props}>
+        {/* Initials are always rendered as the base layer and serve as the
+          graceful fallback when no imageSrc is passed or the image errors.
```
andrewgos: I feel like the comment is slightly inaccurate? Because we still have the SVG fallback as the fallback of these initials being empty

components/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+export const AllSizes: Story = {
+  parameters: {
+    controls: { disable: true },
+    docs: {
+      canvas: { sourceState: 'none' },
+      description: {
+        story:
+          'All six sizes shown without and with a profile image. <br>' +
+          '**Note:** `xs` and `sm` are image-only sizes — initials are not officially supported at these dimensions per the design spec. ' +
+          'In development, a `console.warn` is raised when `size="xs"` or `size="sm"` is rendered without an `imageSrc` ' +
```
andrewgos: Feel free to disagree, but i think this line doesn't need to be explicitly mentioned in storybook? It's just part of code logic mechanics to warn the devs.

```
 In development, a console.warn is raised when size="xs" or size="sm" is rendered without an imageSrc (including the automatic fallback state when an image fails to load)
```

components/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+          '(including the automatic fallback state when an image fails to load). ' +
+          'Use `size="md"` or larger when initials must be shown.',
+      },
+    },
+  },
+  render: () => (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '1rem' }}>
+      <p style={{ margin: 0 }}>Image (all sizes)</p>
+      <div style={{ display: 'flex', alignItems: 'center', gap: '1rem' }}>
```
andrewgos: subjective opinion: but perhaps we can improve the look with a matrix table? so its easier to see which ones are md lg xl etc
```
                     XS      SM.     MD.     LG.    XL.   XXL       
Images               O        O       O      O      O     O

Initials            N/A      N/A     O      O      O      O


```

components/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+      control: { type: 'text' },
+      description:
+        "Accessible description of the image passed to the inner `<img>`. Use the user's full name. Omit when the avatar is purely decorative.",
+      table: {
+        type: { summary: 'string' },
+      },
+    },
+  },
+  args: {
+    size: 'md',
```
andrewgos: i'd suggest we set the default for the storybook as `undefined` here, so we can have a cleaner look of the initial default code

components/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,145 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { expect } from 'storybook/test';
+import { Avatar } from './Avatar';
+
+// Placeholder image used only in story previews.
+// Replace with a real user photo URL in the consuming application.
+const PLACEHOLDER_IMG = 'https://picsum.photos/seed/mds-avatar/96/96';
```
andrewgos: may not be important, but any way we can make this as a local stored image? 🤔  just thinking to make this storybook offline-friendly and less dependency on something external

components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const allowedSizes: AvatarSize[] = ['xs', 'sm', 'md', 'lg', 'xl', 'xxl'];
+
+export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(
+  (
+    { size, imageSrc, alt, initials, className, ...props }: AvatarProps,
+    ref,
+  ) => {
+    const [imgFailed, setImgFailed] = useState(false);
```
andrewgos, line 48: one from AI that is important to address:


>**High:** Image fallback state can get stuck across src changes, causing a valid new image not to render. Once onError sets imgFailed to true, showImage stays false until remount. If parent updates imageSrc from a broken URL to a valid one, the avatar still shows fallback/initials.
**Why this matters:** real user profile updates or retry flows can silently fail to recover image rendering.
**Suggested fix:** reset imgFailed when imageSrc changes (for example with useEffect on imageSrc), or key the img by imageSrc so it remounts with fresh error state.

components/avatar/Avatar.figma.tsx
```diff
@@ -0,0 +1,41 @@
...
+  props: {
+    size: size,
+    initials: figma.string('Initials'),
+  },
+  example: ({ size, initials }) => (
+    <Avatar
+      size={size}
+      initials={initials}
+      imageSrc="https://example.com/user-photo.jpg"
+      aria-label="User name"
```
andrewgos: shouldnt this be `alt` used for the accesibility label?

components/avatar/Avatar.figma.tsx
```diff
@@ -0,0 +1,41 @@
...
+  lg: 'lg',
+  xl: 'xl',
+  xxl: 'xxl',
+});
+
+// Initials type — all sizes
+figma.connect(Avatar, url, {
+  variant: { Type: 'Initials' },
+  props: {
+    size: size,
```
andrewgos: i havent test it on the Figma itself, but what if we select initials with size xs and sm here?

Maybe we need to restrict the options for this particular case? or provide another codeconnect entry for the example of silhouette placeholder?

andrewgos (changes requested):
Nice work!! Very straight forward implementation of Avatar 👍 

Here are several items i noticed that can be improved, also some from the AIs which I thought worth looking into.

components/avatar/avatar.css
Chocolate-lightning: Simple as.

components/avatar/Avatar.stories.tsx
Chocolate-lightning: Sure, good suggestion.

components/avatar/Avatar.figma.tsx
Chocolate-lightning: Good spot!

components/avatar/Avatar.stories.tsx
Chocolate-lightning: Good idea, I agree that it would make the matrix more 'readable'.

components/avatar/Avatar.stories.tsx
Chocolate-lightning: Updated the comment.

components/avatar/Avatar.tsx
Chocolate-lightning: Comment has been updated.

components/avatar/Avatar.stories.tsx
Chocolate-lightning: Created a /assets dir under .storybook/ so we can have story specific reference assets there which will not be bundled in our build pipeline.

Added a SVG example for the Avatar story rather than having a picsum link.

components/avatar/Avatar.figma.tsx
Chocolate-lightning: CodeConnect updated to have different prop maps that should show the Initials set + Size: xs || sm which switches to the fallback SVG.

components/avatar/Avatar.tsx
Chocolate-lightning: Added a `useEffect` call to reset the imgFailed state when we are provided a new imageSrc. Includes a new test for this functionality.

components/avatar/Avatar.tsx
Chocolate-lightning: I have actioned a portion of this point by trimming the Initials values for whitespace.

With regards to invalid initials i.e. symbols + lowercase etc... I think we should hold off on that regard as we leave the prop open for i18n purposes where if we added logic to it, We'd have to manage cases like Turkish i→İ and if we were to limit consumers to ASCII uppercase, then we would have issues with Arabic languages for example.

Chocolate-lightning:
I have added a further fixup commit as I discovered that the `alt` prop was only being respected when the consumer passed an image. The Avatar will retain that functionality but will now also set the aria-label on the parent span in cases of decorative / silhouette fallbacks and Initials only rendering.

components/avatar/Avatar.tsx
andrewgos: works for me, the initial concern was only around the "empty" state values, where these extra symbols just an extra thought out of it 👍

components/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+    alt: 'Jane Doe',
+  },
+} satisfies Meta<typeof Avatar>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    imageSrc: PLACEHOLDER_IMG,
```
andrewgos, line 63: looks like the image is now not displaying in storybook 👀

components/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+                imageSrc={PLACEHOLDER_IMG}
+                initials="JD"
+                alt="Jane Doe"
+              />
+            </div>
+          ))}
+        </div>
+        {/* Initials row */}
+        <div
+          style={{
```
andrewgos: these inline styles can be refactored to be a reusable variable :)

.storybook/assets/story-placeholder.svg
andrewgos, line 1: perhaps we should get another picture? Got me thinking this was the grey white silhouette placeholder picture 😂 or the `placeholder` in the name of the file got me there 🤔

andrewgos (changes requested):
Good find with the `alt` behaviour without an image!

Just a few more items after the fixup commits 🤏

components/avatar/Avatar.stories.tsx
andrewgos: false alarm. restarting storybook got it back working like you said 👌

.storybook/assets/story-placeholder.svg
Chocolate-lightning: As discussed in the chat, we will now have the stock free Jessica image serve as the Avatar example.

components/avatar/Avatar.stories.tsx
Chocolate-lightning: Rejigged the story to define the styles in a reusable way.

andrewgos (dismissed):
Thanks for improving the styling as well as the accessibility. Happy to let this code proceed 🚀

andrewgos (approved):
👍

Issue #264, opened 2026-06-23
fix: Include component assets in build output


Chocolate-lightning:
### Affected area

Component

### Component or token name

Close Button & Favourite Button

### Package version

4.1.0 -> 5.0.0

### Environment

Within LMS

### Steps to reproduce

1. Run `npm pack`
2. Navigate to LMS code
3. Run `npm install <local MDS tarball>`
4. Run `npm update-packages`
5. Include the testing script from the MDL issues for the test bed
6. Include Close & Favourite buttons into the React code in LMS
7. Run `npx grunt react`
8. Navigate to <yourlocalmoodle>/ds.php

### Expected behavior

Close & Favourite buttons contain svg files for icons.

### Actual behavior

Surrounding dom loads and interactions work but there are no icons.

### Accessibility impact

No impact

### Confirmation

- [x] I searched existing issues and did not find a duplicate.
- [x] I can reproduce this issue with the information provided above.

PR #269, opened 2026-06-29
fix(SVG): Resolve missing SVGs in the production build

andrewgos:
## Description

This PR fixes missing SVG icons in production builds by making component CSS image references self-contained at build time and aligning icon rendering with robust mask-based patterns.

**What changed:**

- Added CSS URL inlining in Vite build config for local image assets referenced from component CSS.
- Updated Choicebox checkmark rendering to use a reusable SVG mask asset instead of embedded inline SVG.
- Moved CloseButton SVG into an assets folder and switched rendering to `mask-image`/`-webkit-mask-image` with token-driven color.
- Improved forced-colors handling in Choicebox checked/disabled states for better system-color compatibility.

**Why going with inline SVG content:**
- It matches our current Vite config `build.cssCodeSplit: false`, which outputs one CSS file `dist/index.css` and favors self-contained styling (inlined SVGs) for consumers.
- This reduces the risk of broken relative asset paths in emitted production CSS for downstream consumers.
- It lowers integration overhead for consumers by avoiding additional static-asset copy/config steps.

Tradeoff: Inlining can slightly increase CSS size (especially when the CSS refers to the same SVG file more than once), but we are prioritizing delivery reliability and simpler consumption.

**Why going with `mask-image` instead of `background-image` for single-color icons:**
- `mask-image` keeps SVGs shape-only, while icon color is controlled in CSS via design tokens.
- This could simplifies state styling (default/hover/disabled/selected) because we update color values instead of swapping assets.
- It improves accessibility of forced-colors/high-contrast support by allowing icon color to map to system colors.

Note: AI instructions files have been edited to enforce this.

https://moodle.atlassian.net/browse/MDS-603

## Type of Change

- [ ] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Screenshots/Previews

N/A

## Checklist

- [ ] I have updated, added, and run tests such as JSUnit, Storybook interactions, or fuzzing to prove my fix is effective or that my feature works

## Additional Notes
N/A


Chocolate-lightning (approved):
Kia ora @andrewgos,

Thanks for working on this issue. 

After testing this PR locally by packing this branch as a tarball and installing it on my Moodle LMS test bed with components configured, I can confirm that the icons for the Close & Favourite buttons are now coming across and I do not see a regression from the Choicebox component.

This PR also addresses the issue I raised last week https://github.com/moodlehq/design-system/issues/264 as well as https://moodle.atlassian.net/browse/MDS-603

I think there may be a time down the line where we don't want to convert assets into inline CSS definitions but from your framing in this PR and the build script, This is likely the optimal approach for getting things going.

With regards to the Chromatic changes, I am not seeing any actual changes in the final markup and I think it may just be an erroneous error caused by shifting from a background-image approach to a mask image. I'd like to double check my understanding on that front but I don't see it as a blocking issue here. It is curious that the shift to a mask did not trigger differences in the Choicebox, may that could direct a small look into.

Cheers,

PR #279, opened 2026-07-07
feat(Tooltip): Add new Tooltip component

andrewgos:
## Description

This PR introduces a new Tooltip component designed for accessible, token-driven, RTL-safe usage across Moodle UI.
Key implementation details:

- Added a new Tooltip component with configurable placement (top, bottom, left, right) and variant (dark, light).
- Implemented runtime validation for placement and variant values with graceful fallback to documented defaults:
  - Invalid placement falls back to top.
  - Invalid variant falls back to dark.
- Added accessible trigger-to-tooltip linking:
  - Tooltip bubble id comes from Floating UI context id.
  - If the child is a valid React element, Tooltip clones the trigger and applies accessibility attributes.
  - Existing `aria-describedby` values are preserved and deduplicated when description mode is used.
  - If the trigger does not appear to have an accessible name, Tooltip falls back to aria-label mode.
  - Tooltip bubble renders with `role="tooltip"`.
- Added robust trigger handling:
  - Invalid child usage fails gracefully without throwing.
  - Development logs are emitted for invalid usage patterns.
- Implemented token-based styling and RTL-safe logical sizing:
  - Uses design tokens for color, typography, spacing, and radius.
  - Uses logical inline sizing properties for RTL/LTR compatibility.
  - Includes reduced-motion handling for tooltip transitions.
- Implemented arrow rendering with Floating UI FloatingArrow (SVG path-based), with variant-aware token coloring.
- Implemented always-visible behavior for showcase stories by using Storybook pseudo states (pseudo.hover targeting `.mds-tooltip`) in shared showcaseParameters, so tooltip bubbles render visible for visual review without adding a force-visible CSS modifier or changing runtime Tooltip behavior.
- Uses the dark feedback background token for the default dark tooltip appearance

https://moodle.atlassian.net/browse/MDS-499

## Screenshots/Previews

Screenshot 2026-07-08 at 10 16 16 am

### TypeScript and props

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes

N/A

### DIFF
--- a/components/tooltip/Tooltip.figma.tsx
+++ b/components/tooltip/Tooltip.figma.tsx
@@ -1,0 +1,28 @@
+import figma from '@figma/code-connect';
+import { Button } from '../button';
+import { Tooltip } from './Tooltip';
+
+figma.connect(
+  Tooltip,
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=11558-438',
+  {
+    props: {
+      label: figma.string('Text'),
+      placement: figma.enum('Placement', {
+        Top: 'top',
+        Bottom: 'bottom',
+        Left: 'left',
+        Right: 'right',
+      }),
+      variant: figma.enum('Variant', {
+        Dark: 'dark',
+        Light: 'light',
+      }),
+    },
+    example: ({ label, placement, variant }) => (
+      <Tooltip label={label} placement={placement} variant={variant}>
+        <Button label="Hover me" variant="secondary" />
+      </Tooltip>
+    ),
+  },
+);
--- a/components/tooltip/Tooltip.stories.tsx
+++ b/components/tooltip/Tooltip.stories.tsx
@@ -1,0 +1,274 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import type React from 'react';
+import { expect, userEvent, waitFor, within } from 'storybook/test';
+import { Button } from '../button';
+import { Tooltip } from './Tooltip';
+
+const placements = ['left', 'top', 'bottom', 'right'] as const;
+const variants = ['dark', 'light'] as const;
+
+const showcaseRowCompactStyle: React.CSSProperties = {
+  display: 'flex',
+  gap: '24px',
+  flexWrap: 'wrap',
+};
+
+const showcaseRowStyle: React.CSSProperties = {
+  display: 'flex',
+  gap: '144px',
+  flexWrap: 'wrap',
+};
+
+const showcaseCellStyle: React.CSSProperties = {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+};
+
+const showcaseParameters = {
+  controls: { disable: true },
+  docs: {
+    canvas: { sourceState: 'none' as const },
+  },
+};
+
+const meta = {
+  title: 'Components/Tooltip',
+  component: Tooltip,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
+    children: {
+      table: {
+        disable: true,
+      },
+    },
+    label: {
+      control: { type: 'text' },
+      description: 'Tooltip text. Must be a caller-supplied translated string.',
+      table: {
+        type: { summary: 'string' },
+      },
+    },
+    placement: {
+      control: { type: 'select' },
+      options: ['top', 'bottom', 'left', 'right'],
+      description:
+        'Preferred physical side of the trigger. Auto-flips to avoid viewport edges. In RTL, left and right still mean viewport-left and viewport-right.',
+      table: {
+        type: { summary: 'top | bottom | left | right' },
+        defaultValue: { summary: 'top' },
+      },
+    },
+    variant: {
+      control: { type: 'select' },
+      options: ['dark', 'light'],
+      description: 'Colour mode.',
+      table: {
+        type: { summary: 'dark | light' },
+        defaultValue: { summary: 'dark' },
+      },
+    },
+  },
+  args: {
+    label: 'This is a tooltip',
+    placement: undefined,
+    variant: undefined,
+    children: <Button label="Hover me" variant="secondary" />,
+  },
+  decorators: [
+    (Story) => (
+      // Extra padding ensures the tooltip bubble is fully visible in the story canvas.
+      <div style={{ padding: '64px' }}>
+        <Story />
+      </div>
+    ),
+  ],
+} satisfies Meta<typeof Tooltip>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const page = within(canvasElement.ownerDocument.body);
+    const trigger = canvas.getByRole('button', { name: 'Hover me' });
+    expect(trigger).toBeInTheDocument();
+
+    await userEvent.hover(trigger);
+
+    const tooltip = page.getByRole('tooltip', { hidden: true });
+    await waitFor(() => expect(tooltip).toHaveAttribute('data-open'));
+    expect(tooltip).toHaveTextContent('This is a tooltip');
+  },
+};
+
+/**
+ * Icon-only controls are a common tooltip trigger. Keep the trigger's
+ * accessible name and expose the tooltip as additional description.
+ */
+export const IconOnlyTrigger: Story = {
+  args: {
+    label: 'More actions for this item',
+  },
+  render: (args) => (
+    <Tooltip label={args.label}>
+      <Button
+        label=""
+        aria-label="More options"
+        variant="secondary"
+        startIcon={<i className="fa-solid fa-ellipsis" aria-hidden="true" />}
+      />
+    </Tooltip>
+  ),
+  play: async ({ canvasElement, args }) => {
+    const canvas = within(canvasElement);
+    const page = within(canvasElement.ownerDocument.body);
+    const trigger = canvas.getByRole('button', { name: 'More options' });
+
+    await userEvent.hover(trigger);
+
+    const tooltip = page.getByRole('tooltip', { hidden: true });
+    await waitFor(() => expect(tooltip).toHaveAttribute('data-open'));
+    expect(tooltip).toHaveTextContent(args.label as string);
+
+    const describedBy = trigger.getAttribute('aria-describedby');
+    expect(describedBy).toBeTruthy();
+    expect(describedBy?.split(/\s+/).filter(Boolean)).toContain(tooltip.id);
+    expect(trigger).toHaveAccessibleName('More options');
+  },
+};
+
+/**
+ * Shows all supported visual variants with a fixed placement.
+ */
+export const Variants: Story = {
+  parameters: showcaseParameters,
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+
+    await waitFor(() =>
+      expect(page.getAllByRole('tooltip', { hidden: true })).toHaveLength(
+        variants.length,
+      ),
+    );
+
+    const tooltips = page.getAllByRole('tooltip', { hidden: true });
+    const openTooltips = tooltips.filter((tooltip) =>
+      tooltip.hasAttribute('data-open'),
+    );
+
+    expect(tooltips).toHaveLength(variants.length);
+    expect(openTooltips).toHaveLength(variants.length);
+
+    for (const [index, variant] of variants.entries()) {
+      expect(tooltips[index]).toHaveTextContent(
+        `Tooltip with ${variant} variant`,
+      );
+    }
+  },
+  render: () => (
+    <div style={showcaseRowStyle}>
+      {variants.map((variant) => (
+        <div key={variant} style={showcaseCellStyle}>
+          <Tooltip
+            label={`Tooltip with ${variant} variant`}
+            variant={variant}
+            data-forced-open=""
+          >
+            <Button
+              label={`${variant[0].toUpperCase()}${variant.slice(1)}`}
+              variant="secondary"
+            />
+          </Tooltip>
+        </div>
+      ))}
+    </div>
+  ),
+};
+
+/**
+ * Shows all supported placements with a fixed visual variant.
+ */
+export const Placements: Story = {
+  parameters: showcaseParameters,
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+
+    await waitFor(() =>
+      expect(page.getAllByRole('tooltip', { hidden: true })).toHaveLength(
+        placements.length,
+      ),
+    );
+
+    const tooltips = page.getAllByRole('tooltip', { hidden: true });
+    const openTooltips = tooltips.filter((tooltip) =>
+      tooltip.hasAttribute('data-open'),
+    );
+
+    expect(tooltips).toHaveLength(placements.length);
+    expect(openTooltips).toHaveLength(placements.length);
+
+    for (const [index, placement] of placements.entries()) {
+      expect(tooltips[index]).toHaveTextContent(`Tooltip on the ${placement}`);
+    }
+  },
+  render: () => (
+    <div style={showcaseRowCompactStyle}>
+      {placements.map((placement) => (
+        <div key={placement} style={showcaseCellStyle}>
+          <Tooltip
+            label={`Tooltip on the ${placement}`}
+            placement={placement}
+            data-forced-open=""
+          >
+            <Button
+              label={`${placement[0].toUpperCase()}${placement.slice(1)}`}
+              variant="secondary"
+            />
+          </Tooltip>
+        </div>
+      ))}
+    </div>
+  ),
+};
+
+/** Pressing Escape while open should dismiss the tooltip (WCAG 2.1 SC 1.4.13). */
+export const EscapeDismiss: Story = {
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const page = within(canvasElement.ownerDocument.body);
+    const trigger = canvas.getByRole('button', { name: 'Hover me' });
+    const tooltip = page.getByRole('tooltip', { hidden: true });
+
+    await userEvent.hover(trigger);
+    await waitFor(() => expect(tooltip).toHaveAttribute('data-open'));
+
+    await userEvent.keyboard('{Escape}');
+    await waitFor(() => expect(tooltip).not.toHaveAttribute('data-open'));
+  },
+};
+
+/**
+ * Floating UI placement values are physical, not logical. In RTL, `left` still
+ * means viewport-left, so `right` is used to place on inline-start.
+ */
+export const RightToLeft: Story = {
+  tags: ['test', 'stable'],
+  decorators: [
+    (Story) => (
+      <div dir="rtl" lang="ar" style={{ padding: '64px' }}>
+        <Story />
+      </div>
+    ),
+  ],
+  args: {
+    label: 'هذه تلميحة',
+    placement: 'right',
+    children: <Button label="مرر هنا" variant="secondary" />,
+  },
+};
--- a/components/tooltip/Tooltip.test.tsx
+++ b/components/tooltip/Tooltip.test.tsx
@@ -1,0 +1,329 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { describe, expect, it } from 'vitest';
+import { Button } from '../button';
+import { Tooltip } from './Tooltip';
+
+describe('Tooltip: Unit Test', () => {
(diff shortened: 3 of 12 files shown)


components/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+          </Tooltip>
+        </div>
+      ))}
+    </div>
+  ),
+};
+
+/** RTL layout — directional arrow should mirror correctly via CSS logical properties. */
+export const RightToLeft: Story = {
+  tags: ['test', 'beta'],
```
Chocolate-lightning: Promote this from beta to stable

components/tooltip/Tooltip.tsx
Chocolate-lightning, line 1: WCAG 2.1 SC 1.4.13 (Content on Hover or Focus) requires that content triggered by hover or focus can be dismissed without moving pointer/focus - specifically, pressing Escape must close it. The MSE checklist item 13 ("Tooltip dismisses on Escape key") is marked complete at the design stage, but no Escape handler exists in the React component. The tooltip is purely CSS-driven and has no JS dismiss mechanism at all.

The simplest approach is to track a dismissed state and force visibility to hidden via an inline style or a modifier class: mds-tooltip--dismissed. Reset on mouseenter/focus. Alternatively, this becomes trivial with @floating-ui/react's useDismiss hook (see MDS-610).

components/tooltip/tooltip.css
```diff
@@ -0,0 +1,215 @@
...
+
+@media (prefers-reduced-motion: reduce) {
+  .mds-tooltip__bubble {
+    transition: none;
+  }
+}
+
+/* Show on hover or when a focusable element inside the wrapper has focus */
+.mds-tooltip:hover .mds-tooltip__bubble,
+.mds-tooltip:focus-within .mds-tooltip__bubble {
```
Chocolate-lightning: The CSS uses `.mds-tooltip:focus-within .mds-tooltip__bubble { visibility: visible; }`. :focus-within activates whenever any descendant receives focus, including via mouse click. The ZeroHeight spec says "Tooltips appear on mouseenter and :focus-visible of the trigger element" - i.e. keyboard focus only. Using :focus-within means clicking a button that has a tooltip will cause the tooltip bubble to flash in (and stay) until the button loses focus, which is unexpected and will confuse users.

components/tooltip/tooltip.css
```diff
@@ -0,0 +1,215 @@
...
+  position: relative;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+}
+
+/* ─── Bubble (content box + arrow) ─────────────────────────────────────────── */
+
+.mds-tooltip__bubble {
+  position: absolute;
```
Chocolate-lightning: The `.mds-tooltip__bubble` is `position: absolute` inside the `.mds-tooltip` wrapper div. In a live Moodle page running Bootstrap 5 - this div will be rendered inline wherever the developer places it. Bootstrap components such as `.table-responsive (overflow: auto)`, `.card (overflow: hidden in some themes)`, `.modal-body`, and elements with transform create a new containing block that clips or repositions any absolute descendant. A tooltip inside a table cell, a clipped card, or a transformed element will be cut off or mispositoned - silently, with no fallback.

components/tooltip/tooltip.css
```diff
@@ -0,0 +1,215 @@
...
+  z-index: 1000;
```
Chocolate-lightning: `.mds-tooltip__bubble` has `z-index: 1000`. Bootstrap 5 defines `--bs-tooltip-zindex: 1080` for its own tooltips, and dropdowns share `z-index: 1000`. In a Moodle page where both a Bootstrap dropdown and an MDS tooltip are visible simultaneously, the tooltip bubble can render behind the dropdown backdrop. Bootstrap also has modals at `z-index: 1055` and offcanvas at 1045.

components/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+    ),
+  ],
+} satisfies Meta<typeof Tooltip>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  play: async ({ canvasElement }) => {
```
Chocolate-lightning, line 96: The Storybook play functions only assert the DOM is rendered - they do not simulate hover or keyboard interaction. Per stories-tests.instructions.md, play functions should cover user interactions and visual/a11y state.

components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,106 @@
...
+}: TooltipProps) => {
+  const id = useId();
+
+  let trigger: ReactNode = children;
+
+  if (isValidElement(children)) {
+    const triggerElement = children as ReactElement<
+      HTMLAttributes<HTMLElement>
+    >;
+    const existingDescribedBy = triggerElement.props['aria-describedby'];
```
Chocolate-lightning: The component always wires aria-describedby from trigger to tooltip. ZeroHeight says: "Use aria-label when the tooltip is the only label." For icon-only buttons (the primary intended use case per the ZeroHeight "When to use" section), aria-describedby is wrong - it supplements an existing label but the icon button has no visible text label. Screen readers will announce the button's accessible name first, then the described-by content; if the button has no accessible name beyond the tooltip, aria-label should be used instead. The current API gives callers no way to switch between the two patterns.

components/tooltip/tooltip.css
```diff
@@ -0,0 +1,215 @@
...
+.mds-tooltip {
+  position: relative;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+}
+
+/* ─── Bubble (content box + arrow) ─────────────────────────────────────────── */
+
+.mds-tooltip__bubble {
```
Chocolate-lightning, line 16: MSE checklist item 15 ("Placement is smart — avoids viewport edges") is marked complete in the design stage, but the CSS implementation has no overflow detection or flip logic. A `placement="top"` tooltip on an element near the top of the viewport will render offscreen with no fallback. Either MDS-610 could address this or we will need to handle this flipping ourselves.

components/tooltip/tooltip.css
```diff
@@ -0,0 +1,215 @@
...
+  position: absolute;
+  z-index: 1000;
+
+  /* Hidden until hover/focus-within */
+  visibility: hidden;
+  opacity: 0;
+  pointer-events: none;
+  transition:
```
Chocolate-lightning, line 22: MSE checklist item 16 ("Delay before showing/hiding avoids accidental triggers") is marked complete in design. The implementation has no delay - the tooltip appears instantly on :hover and :focus-within. A small show delay (200–300ms) is standard UX practice to prevent tooltips from appearing during casual mouse traversal over buttons.

MDS-610 could also address this.

components/tooltip/Tooltip.tsx
Chocolate-lightning, line 1: The component has no touchstart/touchend handler. ZeroHeight acknowledges this ("Do not rely on tooltips for essential information on small or touch screens") but do we need to add in some handlers for this maybe?

components/tooltip/tooltip.css
```diff
@@ -0,0 +1,215 @@
...
+  display: block;
+  max-inline-size: 100%;
+  min-inline-size: 0;
+  font-family: var(--mds-font-family-base);
+  font-weight: var(--mds-font-weight-regular);
+  font-size: var(--mds-font-size-paragraph-small);
+  line-height: var(--mds-line-height-paragraph-small);
+  letter-spacing: var(--mds-letter-spacing-default);
+  word-break: normal;
+  overflow-wrap: anywhere;
```
Chocolate-lightning, line 83: Maybe a small comment outling why the `word-break` + `overflow-wrap` combo is needed here.

components/tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,186 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { Tooltip } from './Tooltip';
```
Chocolate-lightning, line 5: Maybe include the button component here similar to the stories to show our button in context.

components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,106 @@
...
+  'left',
+  'right',
+];
+const allowedVariants: TooltipVariant[] = ['dark', 'light'];
+
+export const Tooltip = ({
+  label,
+  placement = 'top',
+  variant = 'dark',
+  children,
```
Chocolate-lightning, line 68: I think we may be a bit 'inside out' with this approach where a tooltip has to wrap around a relevant component to link them. If you have a look at https://getbootstrap.com/docs/5.0/components/tooltips/ you can see they handle it via data-attr values. I think LMS devs might get tripped up if we were to go with the current approach. Maybe MDS-610 could help here as well?

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Great work here overall. There are a few items that I'd like to see addressed and a couple of considerations I think we need to discuss before merging this in.

It was not noted in the design review but I personally find the gap between the trigger and the tooltip a bit large.

components/tooltip/Tooltip.stories.tsx
andrewgos: Updating to stable

components/tooltip/tooltip.css
andrewgos: implementing `@floating-ui/react`, hence this is now resolved

components/tooltip/tooltip.css
andrewgos: resolved along with the `@floating-ui/react`'s portal

components/tooltip/tooltip.css
andrewgos: setting the z-index to be 1080 👍

components/tooltip/tooltip.css
andrewgos: addressed with floatingui's API to add delays, setting to 200ms

components/tooltip/tooltip.css
andrewgos: it is now no longer CSS controlled but using floating UI's API

components/tooltip/tooltip.css
andrewgos: adding a comment 👍

components/tooltip/Tooltip.stories.tsx
andrewgos: updating the play function

components/tooltip/Tooltip.test.tsx
andrewgos: good call. replacing buttons with the Button component here

components/tooltip/Tooltip.tsx
andrewgos: got the `@floating-ui/react` in 👍

andrewgos: added useClick handlers on floating ui's api

components/tooltip/Tooltip.tsx
andrewgos: adding the handling when trigger has no text

components/tooltip/Tooltip.tsx
andrewgos: even with MDS-610 `@floating-ui/react`'s the approach to the trigger still for us to decide.

After consulting further with multiple AIs and checking other design system like [MaterialUI's tooltip](https://mui.com/material-ui/react-tooltip/#basic-tooltip), it is indeed the most 'best' practice for the trigger item being the rendered child, with its own benefits

In terms of the developer experience for LMS devs, I reckon it will be part of getting them to read our documentation and implement it as intended, not letting them just blindly following Bootstrap's doc.

Chocolate-lightning: Yep fair shout, After considering it a bit longer and looking at other design systems and React libs, it does seem that this approach is an accepted standard so I'll resolve this one as a wont do.

components/tooltip/Tooltip.stories.tsx
Chocolate-lightning, line 1: Looking at the ZH docs wrt: "When to use" section,  it leads with an icon only button example but we do not have such a case in our storybook examples. Perhaps we should add one in and then we can also verify the aria handling works as expected as well?

components/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+      description: 'Tooltip text. Must be a caller-supplied translated string.',
+      table: {
+        type: { summary: 'string' },
+      },
+    },
+    placement: {
+      control: { type: 'select' },
+      options: ['top', 'bottom', 'left', 'right'],
+      description:
+        'Preferred side of the trigger. Auto-flips to avoid viewport edges.',
```
Chocolate-lightning: We should probably make it a bit more explicit that even in RTL languages that a position of left will still show on the left, Similar to what you have added to the RTL story.

components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,249 @@
...
+import { cloneElement, isValidElement, useEffect, useState } from 'react';
+import './tooltip.css';
+
+type TooltipPlacement = 'top' | 'bottom' | 'left' | 'right';
+type TooltipVariant = 'dark' | 'light';
+
+export interface TooltipProps extends HTMLAttributes<HTMLDivElement> {
+  /** Tooltip text displayed in the popup. Must be a caller-supplied translated string. */
+  label: string;
+  /** Preferred side of the trigger the tooltip appears on. Defaults to `top`. Auto-flips to avoid viewport edges. */
```
Chocolate-lightning: Extend this documentation to reflect the same info mention in the SB file about how RTL will not shift this property to the inverse.

components/tooltip/Tooltip.figma.tsx
```diff
@@ -0,0 +1,27 @@
+import figma from '@figma/code-connect';
+import { Tooltip } from './Tooltip';
```
Chocolate-lightning, line 3: Can we update this to reference the Button component as well? Similar to the tests and stories.

components/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+  for (const bubble of tooltipBubbles) {
+    bubble.setAttribute('data-open', '');
+    bubble.setAttribute('data-positioned', '');
+    bubble.style.visibility = 'visible';
+    bubble.style.opacity = '1';
+    bubble.style.pointerEvents = 'auto';
+  }
+};
+
+const AlwaysVisibleTooltipShowcase = ({
```
Chocolate-lightning: Maybe we could have a testing CSS rule that sets `visibility: visible; opacity: 1; pointer-events: auto` and then pass `data-forced-open` via the prop attrs. This would probably be a bit less fragile i.e. if we were to change the delay of the open from 200ms then the 220ms delay in this function may break in odd ways. What do you think?

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Thanks for taking my feedback and suggestions on board. This is looking really good now and I only have a couple of minor items that I'd like to see addressed.

Cheers,

components/tooltip/Tooltip.stories.tsx
andrewgos: good call. adding one icon-only story

components/tooltip/Tooltip.stories.tsx
andrewgos: done 👍

components/tooltip/Tooltip.tsx
andrewgos: sure 👍

components/tooltip/Tooltip.figma.tsx
andrewgos: good idea

components/tooltip/Tooltip.stories.tsx
andrewgos: good call.. i initially thought of avoiding altering the API props, but lets make it `data-forced-open` just as an internal prop without changing the exposed public API

andrewgos:
@Chocolate-lightning  Pushed a fixup for the `@floating-ui/react` issue in LMS.

The approach here is to externalize that package, because with `preserveModules:true`, its creating a new `node_modules` within the `dist/` which we do not want. Hence noting `@floating-ui/react` as external, so that the consumer's `node_modules` will include `@floating-ui/react` without explicitly installing one.

Hoooowever, the problem doesn't stop here. LMS has a limitation that it doesn't resolve npm packages well. 
Screenshot 2026-07-14 at 9 26 05 am
LMS might need to do some extra piping, just like the other packages.
Screenshot 2026-07-14 at 9 20 35 am

This work should be part of the MDL task as it updates to the DS release that contains Tooltip.

But from DS's package release perspective, I believe this one is good to go. Let me know what you think!

components/tooltip/Tooltip.test.tsx
Chocolate-lightning, line 1: Looking at some of these tests, whilst valid, could be combined together where test similar or behaviour such as the light, dark & invalid variants. Similarly class & custom props could likely be merged along. The Hover, trigger and escaping control tests could likely benefit from the same treatment to help reduce the total count of tests whilst not losing coverage. Interested to hear your thoughts though.

vite.config.ts
Chocolate-lightning, line 1: I will hold off merging this component until we chat with the LMS teams about how they want us to handle a new dep.

components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,259 @@
...
+  FloatingPortal,
+  arrow,
+  autoUpdate,
+  flip,
+  limitShift,
+  offset,
+  shift,
+  useClick,
+  useDismiss,
+  useFloating,
```
Chocolate-lightning, line 12: `useFloating` is configured with `whileElementsMounted: autoUpdate`. Because the bubble is always mounted in the DOM (via `FloatingPortal`) regardless of open state, `autoUpdate` immediately attaches scroll and resize listeners for every tooltip instance on the page and never unsubscribes until the host component unmounts. For a typical Moodle page with a handful of tooltips this is negligible, but on a dense page (e.g. a table with 50 icon-button cells, each with a tooltip) the listener count could become noticeable.

Maybe consider conditionally passing whileElementsMounted only when the tooltip is open, or using autoUpdate with the animationFrame: false option to reduce callback frequency.

components/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,269 @@
...
+};
+
+const showcaseParameters = {
+  controls: { disable: true },
+  docs: {
+    canvas: { sourceState: 'none' as const },
+  },
+};
+
+const meta = {
```
Chocolate-lightning, line 35: Add `children: { table: { disable: true } }` to the `argTypes` object to hide it from the Controls panel, This is consistent with how other design system components handle non-configurable render-prop children. Currently SB is inferring a value via autodocs which may confuse our users.

components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,259 @@
...
+const ARROW_WIDTH = 12;
+const ARROW_GAP = 8;
+
+export const Tooltip = ({
+  label,
+  placement = 'top',
+  variant = 'dark',
+  children,
+  className,
+  ...props
```
Chocolate-lightning, line 70: `data-forced-open` is read from `props` via a type cast (`(props as Record<string, unknown>)['data-forced-open']`) rather than being destructured. As a result it remains in `...props` and is spread onto the outer `<div className="mds-tooltip" ...>` alongside the bubble. In rendered Storybook stories, the host wrapper div has `data-forced-open=""` on it, which doesn't affect styling (no CSS rule targets `.mds-tooltip[data-forced-open]`) but pollutes the DOM and could mislead a developer inspecting the element tree who may not know the attribute is supposed to be on the bubble only.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

I reckon we are nearly there with a few suggestions left after looking at everything.
Also, do not forget to run through the MSE checklist.

Cheers,

components/tooltip/Tooltip.test.tsx
andrewgos: good point. getting the AI to refactor these tests around 👍

components/tooltip/Tooltip.tsx
andrewgos: fair call, adding `isTooltipOpen` to the `whileElementsMounted`

components/tooltip/Tooltip.tsx
andrewgos: valid situation there 👍 destructuring it away from `wrapperProps`

components/tooltip/Tooltip.stories.tsx
andrewgos: lovely, thanks! i myself was annoyed seeing those noise in the controls panel hahah

Chocolate-lightning (approved):
Kia ora @andrewgos,

Everything is looking good here. CI is passing, local tests are good & running the component on LMS with the library patch included is also passing.

Merging this in.

PR #291, opened 2026-07-15
feat(Switch): Add new Switch component

andrewgos:
## Description

Add a new Switch component built as a checkbox-backed control with semantic variants for `enablement`, `visibility`, and `lock`. The component supports label placement on either side, an optional hidden-label mode with accessible naming fallback, and state styling for hover, checked, disabled, and focus-visible interactions.

Implementation details:

- The control is implemented as a native `<input type="checkbox">` wrapped by a `<label>` so the whole visual surface remains clickable and keyboard-accessible.
- Variant behavior is driven by class names and CSS mask icons, with different icon assets for checked/unchecked states.
- `labelSide` is normalized at runtime to `end` or `start`, and `variant` falls back to `enablement` if an invalid value is passed.
- `hideLabel` keeps the label visually hidden while preserving an accessible name via `aria-label` or the `label` prop.
- The Storybook set includes the default interactive story, variant matrix, label-side showcase, state matrix, and an RTL story.
- Unit tests cover rendering, prop forwarding, controlled/uncontrolled behavior, invalid-value fallback, disabled state, and accessible-name behavior.

Note: **Switch is intentionally made to not support multi-line text**, it is enforced in the CSS and displayed in the storybook. Any text overflow brokenness made as a result will surface to any consumers that they are using it wrong. ZH documentation for more usage info on the Switch: https://design.moodle.com/98292f05f/p/585368-switch

https://moodle.atlassian.net/browse/MDS-556

## Screenshots/Previews
Screenshot 2026-07-16 at 8 56 55 am

### TypeScript and props

- [ ] `...props` spread last on the host element
  - Intentionally not on the last one to be a stricter/safer component on the native checkbox element

### Internationalisation

### CSS and tokens

  - On Figma, the switch intentionally uses a few fixed geometry values for indicator, and thumb sizing, plus SVG mask bounds.

### Tests and stories

### Breaking changes

### AI context

- [ ] Instruction files updated if this component introduces new patterns or anti-patterns
  - No updates done to the instruction files

## Cross-system parity


components/switch/Switch.tsx
```diff
@@ -0,0 +1,137 @@
...
+      : undefined;
+    const shouldRenderVisibleLabel = !hideLabel && !!label;
+
+    return (
+      <div className={classes.filter(Boolean).join(' ')}>
+        <input
+          {...inputProps}
+          id={id}
+          ref={ref}
+          type="checkbox"
```
Chocolate-lightning, line 115: The underlying `<input type="checkbox">` does not carry `role="switch"`. Because the visible indicator is fully hidden from the accessibility tree (`aria-hidden="true"`), the only thing AT sees is the visually-hidden checkbox. Without `role="switch"`, all screen readers announce it as "checkbox", which is semantically incorrect for a binary toggle whose design intent, ZeroHeight documentation, and component API all frame it as a switch. WAI-ARIA 1.1 defines `role="switch"` specifically for this pattern: a checkbox that represents on/off rather than selected/not-selected. This is supported across all modern AT+browser combinations.

You can read more at: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/switch_role & https://www.w3.org/WAI/ARIA/apg/patterns/switch/ along with the reference snippets there.

components/switch/switch.css
Chocolate-lightning, line 1: `inline-size: 14px` / `block-size: 14px` (thumb), `inline-size: 40px` (indicator frame), and `transform: translateX(14px)` (thumb travel) are hardcoded. The project guardrail in `copilot-instructions.md` is explicit: "if no `--mds-*` token exists for a CSS value, don't invent an ad-hoc value — request one from the design team at design.moodle.com".

The track padding already has a token (`--mds-switch-track-padding`), making the absence of tokens for the other geometric values inconsistent. If we were to not have specific geometric tokens for the switch, maybe we can add comments to explain the rationial on not doing so, wdyt?

Chocolate-lightning, line 1: Within other components, we have included a reduced animation query for accessibility. We should likely emulate this pattern within this component to remove the transitions if the user requests it.

components/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,454 @@
...
+      <Switch
+        {...switchArgs}
+        checked={!!checked}
+        onChange={(event) =>
+          updateArgs({ checked: event.currentTarget.checked })
+        }
+      />
+    );
+  },
+  play: async ({ canvas, userEvent }) => {
```
Chocolate-lightning, line 287: The `Default` story's `play` function clicks the switch and then only asserts `toBeVisible()`. This does not verify that the interaction actually toggled the switch state — a regression that silently breaks the checked/unchecked transition would pass this test. The initial `args.checked` is `false`, so after clicking, the expectation should be `toBeChecked()`.

components/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,454 @@
...
+import { type SwitchVariant, Switch } from './Switch';
+
+const meta = {
+  title: 'Components/Switch',
+  component: Switch,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  argTypes: {
```
Chocolate-lightning, line 14: `checked` is listed in `argTypes` with a boolean control, but `onChange` is not. In autodocs, developers see they can pass `checked` but have no guidance that `onChange` is required to make the controlled pattern work. Since `Switch` extends `InputHTMLAttributes<HTMLInputElement>`, `onChange` is part of the public API but is invisible in the generated docs.

components/switch/Switch.test.tsx
Chocolate-lightning, line 1: We are missing fuzz testing for the text input within this component. It should be added to ensure folks can't break out of the component with spurious data.

components/switch/Switch.tsx
```diff
@@ -0,0 +1,137 @@
+import { type InputHTMLAttributes, forwardRef, useId } from 'react';
+
+export type SwitchVariant = 'enablement' | 'visibility' | 'lock';
+export type SwitchLabelSide = 'end' | 'start';
+
+const allowedVariants: SwitchVariant[] = ['enablement', 'visibility', 'lock'];
```
Chocolate-lightning: I am not totally sold on the `enablement` variant name. It is consistent across Figma & ZH but maybe something like `enable` or `toggle` would read better as a `propArg` wdyt?

components/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,454 @@
...
+    await expect(
+      canvas.getByRole('checkbox', { name: 'Visibility enabled unchecked' }),
+    ).not.toBeChecked();
+    await expect(
+      canvas.getByRole('checkbox', { name: 'Lock enabled checked' }),
+    ).toBeChecked();
+  },
+};
+
+export const LabelSides: Story = {
```
Chocolate-lightning, line 374: In this showcase, the `labelSide="end"` and `labelSide="start"` variants have the same DOM widths but appear visually offset, with the label-end variant showing a gap before the indicator. This is likely caused by `.mds-switch-indicator` having `padding: var(--mds-spacing-xxs)` on all sides - the padding creates a buffer that reads as a gap when the indicator is at the leading edge of a flex container. ZeroHeight's layout guidelines say the container should hug its content without extra leading space. Worth verifying against the Figma spec and with a designer.

components/switch/Switch.tsx
```diff
@@ -0,0 +1,137 @@
...
+      'aria-label': ariaLabelProp,
+      disabled,
+      ...inputProps
+    }: SwitchProps,
+    ref,
+  ) => {
+    const generatedId = useId();
+    const id = idProp ?? generatedId;
+
+    if (
```
Chocolate-lightning: Minor improvement: This validation as well as the next one could likely be bundled with the rest of the validation conditions within the overall `if (import.meta.env.DEV) {}` statement, wdyt?

components/switch/Switch.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    it('does not render a visible label element when label is an empty string', () => {
+      const { container } = render(<Switch label="" hideLabel={false} />);
+      expect(
+        container.querySelector('.mds-switch-label'),
+      ).not.toBeInTheDocument();
+    });
+  });
+
+  describe('development warnings', () => {
```
Chocolate-lightning, line 188: This batch of tests fires a render that exercises the fallback, and there are separate tests for fallback class application further up in the file. These test the same render path twice essentially. Combining them into parameterised cases that check both the warning message and the resulting class in a single render would reduce noise in the testing suite.

components/switch/Switch.figma.tsx
```diff
@@ -0,0 +1,51 @@
...
+  variant: figma.enum('Variant', {
+    Enablement: 'enablement',
+    Visibility: 'visibility',
+    Lock: 'lock',
+  }),
+  labelSide: figma.enum('Label side', {
+    End: 'end',
+    Start: 'start',
+  }),
+  checked: figma.enum('Checked', { Yes: true, No: false }),
```
Chocolate-lightning, line 19: Both `figma.connect` calls here include `checked={checked} `driven by `figma.enum('Checked', { Yes: true, No: false })`. The `onChange={() => {}}` stub is correct as a no-op for a static Figma snippet. However, Figma Dev Mode will emit the snippet with a static `checked={true}` or `checked={false}` and a no-op `onChange`, which is potentially misleading for developers copying from Dev Mode who may not realise controlled mode requires a real handler. A brief inline comment in the Code Connect file would help.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Great work on this component. I have a few suggestions and questions for you but overall, it is looking really solid.

One other thing that I noticed was that after running `npm install`, I noticed that the `package-lock.json` file was updated. It may be a local issue for me or a rebasing issue but it is worth double checking it.

Cheers,

components/switch/Switch.tsx
andrewgos: great one. adding it in

components/switch/switch.css
andrewgos: fair call. yes it was intended decision as confirmed by @edumoodle, these hardcoded sizing aren't worth to be a token. from those hardcoded ones it builds up the component using the padding, etc. 

will add a note in the code comments 👍

andrewgos: where/which component do we do that reduced animation query? 🤔 

I'm adding one to the switch 👍

components/switch/Switch.stories.tsx
andrewgos: good spotting, changing to `toBeChecked`

components/switch/Switch.stories.tsx
andrewgos: adding `onChange` !

components/switch/Switch.test.tsx
andrewgos: fuzz tests added 👍

components/switch/Switch.tsx
andrewgos: Thanks for the thoughts!

This was actually something we've rethought in Figma, previously it was `status` changed to `enablement`

Will need your thoughts too @edumoodle

components/switch/Switch.stories.tsx
andrewgos: I can see that its what we have in Figma, with padding on all 4 sides on the indicator.

Can you please confirm if this was intentional @edumoodle , though it is against Zeroheight's layout guidelines?

components/switch/Switch.tsx
andrewgos: sure thing! happy to take that on board

components/switch/Switch.test.tsx
andrewgos: combining those 👍

components/switch/Switch.figma.tsx
andrewgos: adding comments 👍

components/switch/Switch.tsx
andrewgos: As discussed with @edumoodle , we are renaming it to `enable` variant

components/switch/Switch.stories.tsx
andrewgos: The Figma has been updated with new gap and padding, and the code is now matching the latest figma

Thanks @edumoodle !!

components/switch/switch.css
andrewgos: we did similarly to the tooltip 👍  the PR for tooltip hasnt been merged hence i wasnt able to find it haha

Chocolate-lightning (approved):
Kia ora @andrewgos,

Thanks for the rebase & addressing my review items.
Testing against CI, Locally in SB, Unit tests & PVT in LMS all looks great.

The component mirrors the ZH & Figma resources well and CC dryrun is good.

Merging this in. Make sure to update the code tab in ZH.

PR #297, opened 2026-07-20
chore(tokens): Align token generation with the new DTCG format

borisbasque:

## Summary
This PR updates token generation to match the new DTCG format and keeps token relationships clearer in SCSS.

## Why keep one token source file
- The new export now comes as one DTCG file (Light mode), so we keep it in that shape.
- This avoids splitting and recombining token data during build.
- It reduces maintenance risk and preserves the source token graph as provided.
- There is no consumer requirement for per-scope token imports, so keeping a single source file does not remove any required public consumption pattern, keeping the public API the same

## Why SCSS now keeps references
- SCSS now enables output references so aliases stay semantic instead of always flattening to raw values.
- This makes generated SCSS easier to read and maintain, and also aligns SCSS behavior with CSS output.
- Previously, SCSS was flattened because references across separate generated files were unreliable.
- Now with the new single-file token source, cross-file reference issues are reduced, so keeping references is now practical and safer.

## Is This a Breaking Change?

**Short answer: No, by the published public API — but the individual CSS/SCSS files were URL-accessible and their removal is a grey area worth communicating proactively.**

### Why it is not a breaking change

`package.json` exports only three token paths:

| Export | Resolves to |
|---|---|
| `@moodlehq/design-system/tokens/css` | `dist/tokens/css/index.css` |
| `@moodlehq/design-system/tokens/scss` | `dist/tokens/scss/_index.scss` |
| `@moodlehq/design-system/tokens/scss/legacy` | `dist/tokens/scss/_index.legacy.scss` |

All three of these entry points are **unchanged in path and content structure**. No published API contract is broken.

### Why it could be perceived as one

`vite.config.ts` emits **every file** in `tokens/css/` and `tokens/scss/` into `dist/`. Before this PR, `dist/tokens/css/` contained 8 individual category files (`borders.css`, `colors.css`, `primitives.css`, etc.). After this PR, only `index.css` and `light.css` exist there.

Because the project architecture explicitly allows Moodle LMS to load files **directly by URL** (no consumer-side build step), and Vite makes all token files URL-addressable, any consumer that was importing:

```
dist/tokens/css/colors.css
dist/tokens/css/primitives.css
dist/tokens/scss/_typography.scss
```
…directly by URL would get a 404 after this change.

## Other key changes
- Added token name normalization for the new DTCG path structure.
- Added cleanup of generated token directories before rebuild to avoid stale files.
- Updated filename conversion for .tokens.json and Defaultmode naming patterns.
- Regenerated CSS/SCSS outputs to match the new structure.

## Validation
- No Chromatic issues


scripts/tokens.ts
Chocolate-lightning, line 1: Judging by the diff & the PR description, it seems we wont need the aggregator definitions in this file anymore no? or were they being retained for a future 'dark' variant of tokens?

Chocolate-lightning: After thinking about this more, maybe it will be fine to add a comment in `scripts/tokens.ts` near the aggregator definitions explicitly linking to the Jira and explaining that the aggregators are retained in anticipation of a future potential dark-mode token file. This addresses the open review thread and documents the intent for future contributors.

```
/**
 * Custom formatter: css/aggregator
 * Aggregator to create a single CSS file that imports all other generated CSS files.
 * Retained to support future multi-mode token exports (e.g. dark mode) without script changes.
 * See: MDSJPD-3 — https://moodle.atlassian.net/jira/polaris/projects/MDSJPD/ideas/view/11860439
 */
```

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Just a couple of housekeeping items on this one.

Besides the aggregator question, I think that we should have a specific section in the PR description discussing why this is not a breaking change in LMS as I could see it being argued in both directions rather than as a small line item in one of the existing sections.

Such a section could take the shape of the following:

## Is This a Breaking Change?

**Short answer: No, by the published public API — but the individual CSS/SCSS files were URL-accessible and their removal is a grey area worth communicating proactively.**

### Why it is not a breaking change

`package.json` exports only three token paths:

| Export | Resolves to |
|---|---|
| `@moodlehq/design-system/tokens/css` | `dist/tokens/css/index.css` |
| `@moodlehq/design-system/tokens/scss` | `dist/tokens/scss/_index.scss` |
| `@moodlehq/design-system/tokens/scss/legacy` | `dist/tokens/scss/_index.legacy.scss` |

All three of these entry points are **unchanged in path and content structure**. No published API contract is broken.

### Why it could be perceived as one

`vite.config.ts` emits **every file** in `tokens/css/` and `tokens/scss/` into `dist/`. Before this PR, `dist/tokens/css/` contained 8 individual category files (`borders.css`, `colors.css`, `primitives.css`, etc.). After this PR, only `index.css` and `light.css` exist there.

Because the project architecture explicitly allows Moodle LMS to load files **directly by URL** (no consumer-side build step), and Vite makes all token files URL-addressable, any consumer that was importing:

```
dist/tokens/css/colors.css
dist/tokens/css/primitives.css
dist/tokens/scss/_typography.scss
```
…directly by URL would get a 404 after this change.

----

This could then be backed up by having a testing step doing a grep for `design-system.*tokens/css/`

scripts/tokens.ts
andrewgos: That's right, one of the reasons was to anticipate the upcoming dark mode (if any), especially being the default now as `light`

The other reason was not to break the LMS, as the LMS is referencing `_index.legacy` directly, but you are right; by the API contract, we can alter the file name without changing the exposed path. But in this particular case, its clearer/scalable for the LMS to consume `index` instead of `light`

___________

In regards to the note for the PR description, great idea, i'm copy pasting it to the main description 👍 

Also adding the comment for the dark mode 👍

Chocolate-lightning (approved):
Thanks for actioning my feedback. Just squash it down and I'll merge it in.

PR #299, opened 2026-07-20
fix(Radio, Checkbox): Fix alignment issues

andrewgos:
## Description

This PR fixes alignment issues and tightens selector specificity in Radio and Checkbox when rendered alongside Bootstrap form styles.

**Radio — class hooks renamed**
- Root wrapper: `mds-form-check` → `mds-radio`
- Input: `mds-form-check-input` → `mds-radio-input`
- Label: `mds-form-check-label` → `mds-radio-label`
- Feedback: `mds-form-check-feedback` → `mds-radio-feedback`
- Unit tests updated to assert the new class names.

**Radio & Checkbox — input-wrapper pattern**
- Wrapped `<input>` in a `*-input-wrapper` div (flex, `min-height: 24px`) so the indicator enforces the Figma touch target and optically aligns with the first line of a multi-line label.
- All CSS sibling selectors that previously targeted `input ~ label/feedback` updated to use `:has()` on the wrapper (`wrapper:has(.is-invalid) ~ label`) since the input is no longer a direct sibling.
- Added `display: block` to both `mds-checkbox-feedback` and `mds-radio-feedback` — Bootstrap hides `.invalid-feedback` by default via `.is-invalid ~ .invalid-feedback`, which no longer reaches the feedback div across the wrapper boundary.
- Added `WrappedLabel` stories to both `Checkbox.stories.tsx` and `Radio.stories.tsx` to verify indicator top-alignment and text wrapping at narrow widths.

**Radio & Checkbox — selector specificity**
- All CSS rules qualified with `[class]` on the root (e.g. `.mds-radio[class]`, `.mds-checkbox[class]`), raising specificity above Bootstrap's single-class selectors without depending on optional classes like `form-check`.

**Radio & Checkbox — disabled takes precedence over invalid**
- `disabled` destructured in both components and added to the `feedbackId` guard — feedback div is not rendered when `disabled` is true.
- CSS restored: `is-invalid:disabled` state shows muted label colour, not danger.
- New unit tests in `Checkbox.test.tsx` cover the `disabled + invalid` combination.

**Checkbox — CSS cleanup**
- Removed duplicated selectors targeting both `.form-check-label` and `.mds-checkbox-label`; now only `mds-checkbox-label` is targeted.
- Consolidated `:checked` and `:indeterminate` state rules using `:is()`.
- Supporting text colour logic clarified: subtle by default, danger only when `is-invalid:not(:disabled)`.

**Storybook UX improvements (both components)**
- `aria-label` control shown only when `hideLabel` is true (`if: { arg: 'hideLabel', truthy: true }`).
- `invalidFeedback` control shown only when `invalid` is true.
- Default stories converted to controlled args via `useArgs` so Storybook controls stay in sync with rendered state.
- `DefaultChecked` story renamed to "Uncontrolled checked" with an explicit code example.
- `className` argType removed from Checkbox docs (not useful to surface in controls).
- Play functions updated to assert stable visible state without leaving focus styling active.

**Instructions update**
- `.github/instructions/components.instructions.md` — minor formatting fix in the focus-ring CSS example.

**Why `[class]` instead of `.mds-radio.form-check`**

In `hideLabel` mode, `form-check` is intentionally not applied to the wrapper, so `.mds-radio.form-check` selectors would silently stop matching. `[class]` is always present and raises specificity without depending on optional Bootstrap classes.

https://moodle.atlassian.net/browse/MDS-607

## Type of Change

- [ ] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Screenshots/Previews
Screenshots from LMS PVT testbed:
Screenshot 2026-07-21 at 2 31 51 pm
Screenshot 2026-07-21 at 2 30 54 pm

## Checklist

## Additional Notes
- The parent of the Radio button (in the PVT testbed) should now have been adjusted as per our SB docs cc @Chocolate-lightning 
> The two layouts below use a flex container to demonstrate both common arrangements. Neither is provided by this component — choose whichever suits the surrounding form.
```
  display: flex;
  flex-direction: column;
```

### DIFF
--- a/components/checkbox/checkbox.css
+++ b/components/checkbox/checkbox.css
@@ -1,9 +1,9 @@
-.mds-checkbox {
+/* Use [class] to outrank Bootstrap selectors like .form-check .form-check-input */
+.mds-checkbox[class] {
   display: inline-grid;
   grid-template-columns: auto auto;
-  align-items: center;
+  align-items: start;
   align-self: start;
-  min-height: var(--mds-icons-lg);
   padding: var(--mds-spacing-none);
   column-gap: var(--mds-spacing-xs);
   row-gap: var(--mds-spacing-xxs);
@@ -11,17 +11,24 @@
   font-family: var(--mds-font-family-base);
   font-size: var(--mds-font-size-paragraph-default);
   font-weight: var(--mds-font-weight-regular);
-  line-height: var(--mds-line-height-paragraph-xs);
+  line-height: var(--mds-line-height-paragraph-default);
   letter-spacing: var(--mds-letter-spacing-default);
 }
 
-.mds-checkbox-input {
+/* Touch-target wrapper: enforces the Figma minimum height and vertically centres the
+   indicator within it so it optically aligns with the first line of a wrapped label. */
+.mds-checkbox[class] .mds-checkbox-input-wrapper {
+  display: flex;
+  align-items: center;
+  min-height: 24px;
+}
+
+.mds-checkbox[class] .mds-checkbox-input {
   border-radius: var(--mds-border-radius-xs);
   border: var(--mds-stroke-weight-sm) solid
     var(--mds-border-interactive-secondary-default);
-}
-
-.mds-checkbox .mds-checkbox-input {
+  /* Disable Bootstrap's state transitions to avoid a one-frame focus colour flash. */
+  transition: none;
   /* Bootstrap applies float/offset styles to .form-check-input for indented label layouts.
      Match selector specificity and reset float/margins for the grid-based sibling layout. */
   margin: 0;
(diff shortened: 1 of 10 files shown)


components/radio/radio.css
Chocolate-lightning, line 1: I am seeing odd gaps when testing a radio group in LMS, Is this a radio group issue or should margin be added to the radio items themselves? If it is the former, I'll need to update the PVT patch file for LMS. Image

components/radio/radio.css
```diff
@@ -1,133 +1,95 @@
...
-  See .github/instructions/components.instructions.md for full rules and rationale. */
-.mds-form-check-input:focus,
-.mds-form-check-input.is-invalid:focus {
+.mds-radio[class] .mds-radio-input:focus {
   /* Reset Bootstrap's :focus box-shadow; our ring is applied on :focus-visible only */
   box-shadow: none;
   outline: none;
 }
-.mds-form-check-input:focus-visible {
+.mds-radio[class] .mds-radio-input:focus-visible {
```
Chocolate-lightning, line 71: When switching the active radio item within a radio group in LMS, I am seeing odd transitions where it is setting the focusing to black then focus blue, it also fills the radio input and then transitions to the focused style defined by the MDS.

components/checkbox/checkbox.css
```diff
@@ -15,104 +16,105 @@
...
+.mds-checkbox[class] .mds-checkbox-input {
   /* Bootstrap applies float/offset styles to .form-check-input for indented label layouts.
      Match selector specificity and reset float/margins for the grid-based sibling layout. */
   margin: 0;
   float: none;
   background-color: var(--mds-bg-surface-default);
 }
 
-.mds-checkbox-input:checked {
+.mds-checkbox[class] .mds-checkbox-input:checked {
```
Chocolate-lightning: When tabbing and then checking a checkbox in LMS, I am seeing a black focus ring and  then it transitions to focus blue. This should likely be fixed.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Cheers for working on this. I only noticed a couple of issues when testing this in LMS. Please have a look through them and reach out if you have any questions.

components/radio/radio.css
andrewgos: Yes that would be the parent/container issue, just as noted in the Storybook docs, it would be the concern of the parent layout.

> The two layouts below use a flex container to demonstrate both common arrangements. Neither is provided by this component — choose whichever suits the surrounding form.

Correct, the PVT will need to be updated, something like how the SB did it

```
  display: flex;
  flex-direction: column;
  gap: 0.5rem;
```

components/radio/radio.css
andrewgos: eagle eye!!! fixed it with `transition:none`

components/checkbox/checkbox.css
andrewgos: eagle eye!!! fixed it with `transition:none`

Chocolate-lightning (dismissed):
Kia ora @andrewgos,

Thanks for actioning my raised items.

Once you rebase this branch, let me know and I'll merge it.

Chocolate-lightning (approved):
Kia ora @andrewgos,

Thanks for working on this. I have tested this across CI, Local and LMS PVT instances and it is looking good.

I did sneak in a small change into your last commit to address some SB consistency issues and a couple of small doc changes. Nothing major so I'll bypass the rules and merge this in to save you having to make the minor tweaks.

PR #305, opened 2026-07-27
feat(Breadcrumb): Introduce new component

Chocolate-lightning:
## Description

Add the breadcrumb component based on the tooltip & dropdown components.

https://moodle.atlassian.net/browse/MDS-533

### TypeScript and props

### Internationalisation

- [ ] CSS uses logical properties (`margin-inline-*`, `padding-inline-*`, `inset-inline-*`) for any directional layout

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

### DIFF
--- a/components/breadcrumb/Breadcrumb.figma.tsx
+++ b/components/breadcrumb/Breadcrumb.figma.tsx
@@ -1,0 +1,99 @@
+import figma from '@figma/code-connect';
+import { Breadcrumb } from './Breadcrumb';
+
+// NOTE: This file uses the main Moodle Design System file key (bPRkRtSszcbWw9f9p9rXvA)
+// with the node created on the MDS-533-Breadcrumb branch. Update this URL once the
+// branch is merged into the main file.
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=12126-23';
+
+// Items: 2
+figma.connect(Breadcrumb, url, {
+  variant: { Items: '2' },
+  example: () => (
+    <Breadcrumb
+      items={[{ label: 'Home', href: '/' }, { label: 'Current page' }]}
+      ariaLabel="Breadcrumb"
+    />
+  ),
+});
+
+// Items: 3
+figma.connect(Breadcrumb, url, {
+  variant: { Items: '3' },
+  example: () => (
+    <Breadcrumb
+      items={[
+        { label: 'Home', href: '/' },
+        { label: 'Section', href: '/section' },
+        { label: 'Current page' },
+      ]}
+      ariaLabel="Breadcrumb"
+    />
+  ),
+});
+
+// Items: 4
+figma.connect(Breadcrumb, url, {
+  variant: { Items: '4' },
+  example: () => (
+    <Breadcrumb
+      items={[
+        { label: 'Home', href: '/' },
+        { label: 'Section', href: '/section' },
+        { label: 'Sub-section', href: '/sub' },
+        { label: 'Current page' },
+      ]}
+      ariaLabel="Breadcrumb"
+    />
+  ),
+});
+
+// Items: More than 4
+figma.connect(Breadcrumb, url, {
+  variant: { Items: 'More than 4' },
+  example: () => (
+    <Breadcrumb
+      items={[
+        { label: 'Home', href: '/' },
+        { label: 'Level 2', href: '/l2' },
+        { label: 'Level 3', href: '/l3' },
+        { label: 'Level 4', href: '/l4' },
+        { label: 'Level 5', href: '/l5' },
+        { label: 'Current page' },
+      ]}
+      ariaLabel="Breadcrumb"
+    />
+  ),
+});
+
+// Items: More than 4 (long labels / truncation)
+figma.connect(Breadcrumb, url, {
+  variant: { Items: 'More than 4' },
+  example: () => (
+    <Breadcrumb
+      items={[
+        {
+          label: 'Moodle learning platform home page',
+          href: '/',
+        },
+        {
+          label: 'Faculty of Arts and Social Sciences',
+          href: '/faculty',
+        },
+        {
+          label: 'Bachelor of Design and Technology - 2025 intake cohort',
+          href: '/program',
+        },
+        {
+          label: 'DSGN 101 - Introduction to design thinking and process',
+          href: '/course',
+        },
+        {
+          label: 'Week 1 - Foundations and context setting workshop',
+        },
+      ]}
+      ariaLabel="Breadcrumb"
+    />
+  ),
+});
--- a/components/breadcrumb/Breadcrumb.stories.tsx
+++ b/components/breadcrumb/Breadcrumb.stories.tsx
@@ -1,0 +1,237 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { expect, screen, userEvent, within } from 'storybook/test';
+import { Breadcrumb, type BreadcrumbItem } from './Breadcrumb';
+
+// ─── Sample data ──────────────────────────────────────────────────────────────
+
+const sevenItems: BreadcrumbItem[] = [
+  { label: 'Home', href: '#home' },
+  { label: 'Faculty of Arts', href: '#arts' },
+  { label: 'Design & Technology', href: '#dt' },
+  { label: 'Undergraduate', href: '#ug' },
+  { label: 'My courses', href: '#courses' },
+  { label: 'Introduction to design', href: '#course' },
+  { label: 'Week 1: Foundations' },
+];
+
+const shortOverflowItems: BreadcrumbItem[] = [
+  { label: 'Home', href: '#home' },
+  { label: 'Area', href: '#area' },
+  { label: 'Course', href: '#course' },
+  { label: 'Topic', href: '#topic' },
+  { label: 'Week 1' },
+];
+
+const longLabelItems: BreadcrumbItem[] = [
+  { label: 'Home', href: '#home' },
+  {
+    label: 'Faculty of Arts and Social Sciences',
+    href: '#faculty',
+  },
+  {
+    label: 'Bachelor of Design and Technology - 2025 intake',
+    href: '#program',
+  },
+  { label: 'DSGN 101 - Introduction to design thinking and process' },
+];
+
+const rtlSevenItems: BreadcrumbItem[] = [
+  { label: 'الرئيسية', href: '#home' },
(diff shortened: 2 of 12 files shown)


components/breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+      description: {
+        component:
+          "Navigation landmark that shows the user's position in the site hierarchy. " +
+          'Renders up to four items inline; when more are supplied the middle ancestors ' +
+          'collapse behind an overflow button. Link labels are automatically truncated ' +
+          'based on the viewport width, widening progressively across the MDS breakpoints.',
+      },
+    },
+  },
+  tags: ['autodocs', 'test', 'beta'],
```
andrewgos: reminder to update to stable

components/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,242 @@
...
+            .concat('mds-breadcrumb__item--current')
+            .join(' ')}
+          aria-current="page"
+        >
+          <TruncatingTooltip
+            label={currentItem.label}
+            placement="bottom"
+            labelSelector=".mds-breadcrumb__label"
+          >
+            <span className="mds-breadcrumb__label" tabIndex={0}>
```
andrewgos: its recommended to only be keyboard-focusable (tabindex=0) when it is truncated / tooltip rendered, otherwise it'll be misleading while it's not interactable

components/breadcrumb/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+      render(<Breadcrumb items={fourItems} />);
+      expect(screen.getByRole('link', { name: 'Home' })).toBeInTheDocument();
+      expect(screen.getByRole('link', { name: 'Section' })).toBeInTheDocument();
+      expect(
+        screen.getByRole('link', { name: 'Sub-section' }),
+      ).toBeInTheDocument();
+    });
+
+    // Need to confirm if this is the best place to test this or maybe in storybook?
+    it.skip('always wraps link labels with a Tooltip for truncation accessibility', () => {
```
andrewgos: any reason of why is this skipped?

components/breadcrumb/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+      const user = userEvent.setup();
+      render(<Breadcrumb items={sevenItems} />);
+      const trigger = screen.getByRole('button');
+      await user.click(trigger);
+      expect(screen.getByRole('menu')).toBeInTheDocument();
+      await user.keyboard('{Escape}');
+      expect(screen.queryByRole('menu')).not.toBeInTheDocument();
+    });
+
+    it.skip('closes the overflow menu when a menu item is clicked', async () => {
```
andrewgos: any reason of why is this skipped?

components/breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+type Story = StoryObj<typeof meta>;
+
+// ─── Stories ──────────────────────────────────────────────────────────────────
+
+/** The default interactive story. Use the controls panel to change `items`,
+ * `ariaLabel`, and `overflowAriaLabel` and see the result live.
+ */
+export const Default: Story = {
+  args: {
+    items: longLabelItems,
```
andrewgos, line 126: could be just an unpopular opinion: but i think we can leave the Default with the short labels (simpler first impression), while having additional story to demonstrate how truncation works with longlabel items

components/breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+  },
+};
+
+/**
+ * RTL layout with overflow, using Arabic labels. Verifies that separators,
+ * padding, and the overflow dropdown all mirror correctly under `dir="rtl"`.
+ * If overflow works correctly in RTL the simpler four-item case is implicitly
+ * covered.
+ */
+export const RightToLeftOverflow: Story = {
```
andrewgos, line 221: in RTL, i think the dropdown should be aligned to the right of the box instead of the left? 

Image

components/breadcrumb/breadcrumb.css
```diff
@@ -0,0 +1,214 @@
+/*
+ * Override Bootstrap's breadcrumb CSS variables at the component root so the
+ * MDS tokens control all visual output. Bootstrap's structural classes
+ * (breadcrumb, breadcrumb-item, active) are kept for HTML compatibility -
+ * they give Moodle themes a familiar hook - but all colours, spacing, and
+ * typography are driven by --mds-* custom properties.
+ */
+.mds-breadcrumb {
+  --bs-breadcrumb-divider-color: var(--mds-text-subtle);
```
andrewgos: i believe we can do this without having to override bootstrap / define variables like this? otherwise this creates dependencies on the Bootstrap definition

components/breadcrumb/Breadcrumb.figma.tsx
```diff
@@ -0,0 +1,68 @@
+import figma from '@figma/code-connect';
+import { Breadcrumb } from './Breadcrumb';
+
+// NOTE: This file uses the main Moodle Design System file key (bPRkRtSszcbWw9f9p9rXvA)
+// with the node created on the MDS-533-Breadcrumb branch. Update this URL once the
+// branch is merged into the main file.
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=12126-23';
+
```
andrewgos, line 9: maybe we can add another variant to display how truncation works with long labels?

components/breadcrumb/Breadcrumb.stories.tsx
andrewgos: or..... actually leave this crowded so dev can play with the content (label/href) without having to worry about the JSON formatting.

because i find the editting the JSON value is quite frustrating when you dont know the JSON structure haha

in that case, maybe add more child so that the dropdown shows up?

unless we want to create fully-custom playground controls 🤪

andrewgos (changes requested):
thanks for this! definitely easier to digest than the dropdown lol

overall looking neat already, just a few minor items to address :)

components/breadcrumb/Breadcrumb.stories.tsx
Chocolate-lightning: Simple as.

components/breadcrumb/Breadcrumb.test.tsx
Chocolate-lightning: That was left in there as a development artefact since it was throwing constant issues as I was working on the breadcrumbs and was going to revisit later. Actioned.

components/breadcrumb/Breadcrumb.test.tsx
Chocolate-lightning: That was left in there as a development artefact since it was throwing constant issues as I was working on the breadcrumbs and was going to revisit later. Actioned.

components/breadcrumb/Breadcrumb.stories.tsx
Chocolate-lightning: Good spot, Addressed this with the help of the latest dropdown component code.

components/breadcrumb/Breadcrumb.figma.tsx
Chocolate-lightning: Long label + overflow Figma CC link setup.

components/breadcrumb/breadcrumb.css
Chocolate-lightning: Removed the BS overrides.

components/breadcrumb/Breadcrumb.stories.tsx
Chocolate-lightning: Moved to short labels, also added a bit more documentation on the JSON structure to help consumers.

components/breadcrumb/Breadcrumb.tsx
Chocolate-lightning: The current page item is now focusable when it is being actively truncated.

components/breadcrumb/breadcrumb.css
```diff
@@ -0,0 +1,196 @@
...
+  padding: 0;
+  float: none;
+}
+
+.mds-breadcrumb .breadcrumb-item + .breadcrumb-item {
+  /* Neutralize Bootstrap's physical padding so RTL/LTR both rely on the
+     symmetric divider margins above. */
+  padding-left: 0;
+  padding-right: 0;
+  padding-inline-start: 0;
```
andrewgos: these 3 lines 
```
  padding-left: 0;    /* ← physical, remove */
  padding-right: 0;   /* ← physical, remove */
  padding-inline-start: 0;
```
can be simplified as
```
  padding-inline: 0;
```

components/breadcrumb/breadcrumb.css
```diff
@@ -0,0 +1,196 @@
...
+  margin-bottom: 0;
+  padding: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+/* ─── Items ─────────────────────────────────────────────────────────────────── */
+
+.breadcrumb-item.items-2 {
```
andrewgos, line 17: i wasnt able to get the my local PVT working with floating-ui yet. in the meantime, i'll post it here.

Does this needs a better specifity? for example `.mbs-breadcrumb .breadcrumb-item.items-2 {`

may need to check the other bootstrap classes we are overriding in this file too

components/breadcrumb/Breadcrumb.stories.tsx
andrewgos: on top of the content, shouldnt the `...` be on the right of the box? its still aligned to the left of the box

Image

andrewgos (changes requested):
Just few code improvement from the previous changes and looks like RTL still needs some fixes 🙏

components/breadcrumb/breadcrumb.css
Chocolate-lightning: Fair call, updated.

components/breadcrumb/breadcrumb.css
Chocolate-lightning: Good suggestion to check this. Looking at the LMS with the breadcrumbs in LMS, the widths are being applied correctly.

andrewgos (approved):
Thanks for fixing the unwanted dropdown flipping behaviour! LGTM 🚀

PR #318, opened 2026-08-18
feat(Alert): Add new Alert component

andrewgos:
## Description

Introduces the new `Alert` component — a contextual feedback banner for surfacing system messages, errors, warnings, and success states to users. The component renders a semantic live region (`role="status"` or `role="alert"` depending on severity), an icon, optional title, message body, a dismiss control, an inline action button, and a custom content slot.

Key behaviours:
- Four semantic types: `info`, `success`, `warning`, `danger`; falls back gracefully to `info` for invalid values
- Dismissible by default — focus is moved to the next focusable element after the alert is removed from the DOM
- Optional inline action button rendered via the existing `Button` component
- Optional custom content slot (`children`) for arbitrary markup
- `role="alert"` for `warning`/`danger`; `role="status"` for `info`/`success` (ARIA live region parity)
- All user-facing strings accepted as props — no hardcoded text

https://moodle.atlassian.net/browse/MDS-491

## Screenshots/Previews

Screenshot 2026-08-18 at 3 23 32 pm

- [x] All 5 required files present: `Alert.tsx`, `alert.css`, `Alert.test.tsx`, `Alert.stories.tsx`, `index.tsx`
- [x] `Alert.figma.tsx` Code Connect file created and published

### TypeScript and props

- [x] Props interface extends the correct React HTML element interface (`HTMLAttributes<HTMLDivElement>`)

### Internationalisation

### CSS and tokens

### Tests and stories

### Breaking changes

- [x] Breaking change assessment complete — new component, no existing props removed or renamed

### AI context

- [x] `component-index.instructions.md` updated to include Alert entry

## Cross-system parity

## Additional Notes

Focus management on dismiss: after the alert is removed, focus moves to the next focusable element in document order (skipping elements inside the alert). This is handled via `requestAnimationFrame` to ensure the DOM has updated before `.focus()` is called.


components/alert/Alert.tsx
```diff
@@ -0,0 +1,216 @@
...
+    type && allowedTypes.includes(type as AlertType) ? type : 'info';
+
+  if (import.meta.env.DEV) {
+    if (type && !allowedTypes.includes(type as AlertType)) {
+      console.warn(
+        `[MDS Alert] Invalid type "${type}". Falling back to "info". Allowed: ${allowedTypes.join(', ')}`,
+      );
+    }
+
+    if (isDismissible && !dismissAriaLabel?.trim()) {
```
Chocolate-lightning, line 131: I think we should either enforce the label being required in the default state because at the moment we have a default implementation that can basically fail if an optional param is not set which is a bit odd. Anther option would be to flip the default isDismissible to false so that case does not happen. wdyt?

components/alert/Alert.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    );
+
+    const dismissButton = screen.getByRole('button', { name: 'Close alert' });
+    fireEvent.click(dismissButton);
+
+    expect(onDismiss).toHaveBeenCalledOnce();
+    expect(screen.queryByText('Message')).not.toBeInTheDocument();
+  });
+
+  it('moves focus to the next focusable element after dismiss', () => {
```
Chocolate-lightning, line 132: Perhaps we need another test similar to this one that checks what happens if the next focusable element is not found? It is a bit of a non-issue since alerts should never be the last item in the DOM but you never know haha. wdyt?

components/alert/Alert.tsx
```diff
@@ -0,0 +1,216 @@
...
+        nextFocusableElement.focus();
+      });
+    }
+  };
+
+  if (dismissed) {
+    return null;
+  }
+
+  const classes = ['mds-alert', `mds-alert--${resolvedType}`];
```
Chocolate-lightning: Bootstrap Alert classes are not being applied here which fails against MSE item: 17.5, Was there a reason not to include the class hooks?

components/alert/Alert.figma.tsx
```diff
@@ -0,0 +1,232 @@
+import figma from '@figma/code-connect';
+import { Alert } from './Alert';
+
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=13661-871';
+
+// Shared props that apply to every variant combination.
```
Chocolate-lightning, line 7: When validating the Code Connect for this component, I am getting the following error message: `ParserError: Could not find prop mapping for isDismissible ? 'Dismiss alert' : undefined`

components/alert/Alert.tsx
```diff
@@ -0,0 +1,216 @@
...
+  type?: AlertType;
+
+  /** Optional heading above the message body. */
+  title?: string;
+
+  /** Required alert message content. */
+  message: string;
+
+  /** Shows a dismiss control that removes the alert from the page. Defaults to true. */
+  isDismissible?: boolean;
```
Chocolate-lightning, line 20: ZeroHeight and the MSE checklist use `isDismissable` in their written guidance, while Figma, code, Storybook, and the component index use `isDismissible`. This means that we can not mark off 'cross-system parity' items within the MSE

components/alert/Alert.stories.tsx
```diff
@@ -0,0 +1,318 @@
...
+  fontWeight: 'var(--mds-font-weight-medium)',
+  lineHeight: 'var(--mds-line-height-paragraph-default)',
+  color: 'var(--mds-text-default)',
+} as const;
+
+const standardStoryWidthStyle = {
+  inlineSize: '32rem',
+} as const;
+
+export const Default: Story = {
```
Chocolate-lightning, line 156: The targeted stories pass, but the stories do not exercise dismissing, keyboard activation of the close button, action callback behaviour, focus restoration, or the no-following-target case. The unit tests cover some of these behaviours, but the MSE expects state and interaction coverage (States 5.1-5.6, Storybook Stories 19.1-19.4).

components/index.css
```diff
@@ -1,4 +1,5 @@
 @import './activity-icon/activity-icon.css';
+@import './alert/alert.css';
 @import './breadcrumb/breadcrumb.css';
 @import './avatar/avatar.css';
```
Chocolate-lightning, line 3: Not directly related but maybe we could shift this to be before the breadcrumb to restore alphabetical order, wdyt?

components/alert/Alert.test.tsx
```diff
@@ -0,0 +1,145 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { Alert } from './Alert';
+
+describe('Alert: Unit Test', () => {
```
Chocolate-lightning, line 6: Pending the Bootstrap hook discussion, We may need specific tests to cover coexistence between Bootstrap behaviours and our Alert component.

Chocolate-lightning (changes requested):

components/alert/Alert.tsx
andrewgos: fair call on that edge case. having the `isDismissible` was a concious decision to made it on by default.

another approach i think would be suitable is to do a default value on `dismissAriaLabel` in case nothing was provided, just like the other components (Breadcrumb and Pagination)

let me know if you think should be otherwise!

components/alert/Alert.test.tsx
andrewgos: valid point 👍  adding it in

components/alert/Alert.tsx
andrewgos: was unintentional! adding it in.

components/alert/Alert.figma.tsx
andrewgos: fixed with static string literals

components/index.css
andrewgos: ah good catch! its the breadcrumb that needs to move after badge. moved 👍

components/alert/Alert.tsx
andrewgos: `isDismissable` was a typo. ive fixed those in the ZH

components/alert/Alert.stories.tsx
andrewgos: fair call.

extending several play function to be more intentional in testing. added new story FocusRestoredAfterDismiss too

components/alert/Alert.test.tsx
andrewgos: tests handled for the boostrap hooks 👍

components/alert/Alert.tsx
Chocolate-lightning: Make sure to update the MSE as well.

components/alert/Alert.tsx
Chocolate-lightning: I think we can run with the current solution as a safety net.

Chocolate-lightning (approved):

PR #321, opened 2026-08-21
feat(Textarea): Add new Textarea component

andrewgos:
## Description

Adds a new `Textarea` component with accessible labeling, validation feedback, supporting text, optional character counter, optional info tooltip, and Storybook plus unit test coverage for core behaviors.

### Key implementation decisions

- Built on native `<textarea>` semantics plus Bootstrap’s `form-control` baseline, then layered MDS styling via `mds-*` classes and tokenized CSS variables.
- Kept label visibility and accessible naming decoupled: when visual label is hidden, the control is still named through `aria-label` fallback logic.
- Chose conditional invalid feedback rendering (instead of always-rendered hidden message) to keep spoken output concise and avoid unnecessary `aria-describedby` noise.
- Added invalid-state icon so error state is not communicated by color alone.
- Counter is opt-in and only active with `maxLength`; implemented milestone-based live announcements to reduce screen-reader verbosity.
- Counter and supporting/feedback content share a single footer row to preserve layout consistency across states.
- Included RTL coverage in stories and forced LTR numeric counter direction for predictable number formatting in mixed-direction UIs.
- Reused existing `Button` + `Tooltip` for the info affordance to avoid introducing new interaction primitives.

https://moodle.atlassian.net/browse/MDS-641

## Screenshots/Previews
Screenshot 2026-08-21 at 3 22 56 pm

### TypeScript and props

  - Reason: This Textarea API does not introduce constrained string-union props that need runtime narrowing. Existing defaults are handled directly (for example, booleans and rows), so the specific allowedValues pattern is not applied.

### Internationalisation

  - Reason: Some user-facing text is currently hardcoded in component logic, including development warnings and counter/live-region announcement phrases (for example, “of”, “characters remaining”, “over the character limit”), so not all visible/announced strings are externally provided as props.
  - Cleared: New props added to supply strings.

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes
N/A

### DIFF
--- a/components/textarea/Textarea.figma.tsx
+++ b/components/textarea/Textarea.figma.tsx
@@ -1,0 +1,136 @@
+import figma from '@figma/code-connect';
+import { Textarea } from './Textarea';
+
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=13934-4878';
+
+const baseProps = {
+  label: figma.string('Supporting text#13940:21'),
+  required: figma.boolean('Required#13940:15'),
+  // "Show Label" is the inverse of hideLabel
+  hideLabel: figma.boolean('Show Label#13940:13', { true: false, false: true }),
+  resizable: figma.boolean('Resizable#13963:21'),
+  infoTooltipLabel: figma.boolean('Info#13950:34', {
+    true: 'Additional information about this field',
+    false: undefined,
+  }),
+};
+
+// Default state
+figma.connect(Textarea, url, {
+  variant: { State: 'default' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, resizable, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      resizable={resizable}
+      infoTooltipLabel={infoTooltipLabel}
+      placeholder="Placeholder text goes here"
+    />
+  ),
+});
+
+// Default state with supporting text
+figma.connect(Textarea, url, {
+  variant: { State: 'default', 'Support text': 'yes' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, resizable, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      resizable={resizable}
+      infoTooltipLabel={infoTooltipLabel}
+      placeholder="Placeholder text goes here"
+      supportingText="Supporting text"
+    />
+  ),
+});
+
+// Default state with counter
+figma.connect(Textarea, url, {
+  variant: { State: 'default', Counter: 'yes' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, resizable, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      resizable={resizable}
+      infoTooltipLabel={infoTooltipLabel}
+      placeholder="Placeholder text goes here"
+      showCounter
+      maxLength={100}
+    />
+  ),
+});
+
+// Invalid state without feedback text
+figma.connect(Textarea, url, {
+  variant: { State: 'invalid', 'Support text': 'no' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, resizable, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      resizable={resizable}
+      infoTooltipLabel={infoTooltipLabel}
+      placeholder="Placeholder text goes here"
+      invalid
+    />
+  ),
+});
+
+// Invalid state with feedback text
+figma.connect(Textarea, url, {
+  variant: { State: 'invalid', 'Support text': 'yes' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, resizable, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      resizable={resizable}
+      infoTooltipLabel={infoTooltipLabel}
+      placeholder="Placeholder text goes here"
+      invalid
+      invalidFeedback="Error message"
+    />
+  ),
+});
+
+// Disabled state
+figma.connect(Textarea, url, {
+  variant: { State: 'disabled' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      infoTooltipLabel={infoTooltipLabel}
+      placeholder="Placeholder text goes here"
+      disabled
+    />
+  ),
+});
+
+// Read-only state
+figma.connect(Textarea, url, {
+  variant: { State: 'read-only' },
+  props: baseProps,
+  example: ({ label, hideLabel, required, resizable, infoTooltipLabel }) => (
+    <Textarea
+      label={label}
+      hideLabel={hideLabel}
+      required={required}
+      resizable={resizable}
+      infoTooltipLabel={infoTooltipLabel}
+      readOnly
+      defaultValue="Lorem ipsum dolor sit amet"
+    />
+  ),
+});
--- a/components/textarea/Textarea.stories.tsx
+++ b/components/textarea/Textarea.stories.tsx
@@ -1,0 +1,391 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { useState } from 'react';
+import { expect, userEvent, within } from 'storybook/test';
+import { Textarea } from './Textarea';
+
+const meta = {
+  title: 'Components/Textarea',
+  component: Textarea,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs', 'test', 'stable'],
+  decorators: [
+    (Story) => (
+      <div style={{ width: 'min(395px, 95vw)' }}>
+        <Story />
+      </div>
+    ),
+  ],
+  argTypes: {
+    placeholder: {
+      description: 'Placeholder text shown when the field is empty.',
+      control: { type: 'text' },
+    },
+    label: {
+      description: 'Visible label text above the field.',
+      control: { type: 'text' },
+    },
+    hideLabel: {
+      description:
+        'When true, hides the visible label. Provide an accessible name via aria-label or label prop.',
+      control: { type: 'boolean' },
+      table: { defaultValue: { summary: 'false' } },
+    },
+    required: {
+      description: 'Marks the field as required. Appends a * to the label.',
+      control: { type: 'boolean' },
+      table: { defaultValue: { summary: 'false' } },
+    },
+    invalid: {
+      description:
+        'Marks the field as invalid — applies danger border styling and sets aria-invalid.',
+      control: { type: 'boolean' },
+      table: { defaultValue: { summary: 'false' } },
+    },
+    invalidFeedback: {
+      description:
+        'Pre-translated error message shown below the field when invalid is true.',
+      control: { type: 'text' },
+      if: { arg: 'invalid', truthy: true },
+    },
+    supportingText: {
+      description: 'Helper text shown below the field in non-error state.',
+      control: { type: 'text' },
+    },
+    infoTooltipLabel: {
+      description:
+        'Accessible label for the info tooltip button shown beside the visible label.',
+      control: { type: 'text' },
+      if: { arg: 'hideLabel', eq: false },
+    },
+    showCounter: {
+      description:
+        'Shows a live character counter. Requires maxLength to be set.',
+      control: { type: 'boolean' },
+      table: { defaultValue: { summary: 'false' } },
+      if: { arg: 'maxLength', exists: true },
+    },
+    counterAriaLabel: {
+      description:
+        'Optional i18n text (static string) or formatter for the counter aria-label. Formatter receives (currentLength, maxLength). The control only accepts a static string; use the args/code editor to pass a formatter function.',
+      control: { type: 'text' },
+      table: {
+        type: {
+          summary:
+            'string | ((currentLength: number, maxLength: number) => string)',
+        },
+      },
+    },
+    counterRemainingAnnouncement: {
+      description:
+        'Optional i18n text (static string) or formatter for remaining-character milestone announcements. Formatter receives (remaining, maxLength). The control only accepts a static string; use the args/code editor to pass a formatter function.',
+      control: { type: 'text' },
+      table: {
+        type: {
+          summary:
+            'string | ((remaining: number, maxLength: number) => string)',
+        },
+      },
+    },
+    counterOverLimitAnnouncement: {
+      description:
+        'Optional i18n text (static string) or formatter for over-limit announcements. Formatter receives (overLimitBy, maxLength). The control only accepts a static string; use the args/code editor to pass a formatter function.',
+      control: { type: 'text' },
+      table: {
+        type: {
+          summary:
+            'string | ((overLimitBy: number, maxLength: number) => string)',
+        },
+      },
+    },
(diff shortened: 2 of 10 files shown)


components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,304 @@
...
+    const generatedId = useId();
+    const id = idProp ?? generatedId;
+    const isInvalid = !!invalid;
+    const hasVisibleLabel = !hideLabel;
+    const ariaLabel = hideLabel ? (ariaLabelProp ?? label) : undefined;
+
+    // Invalid feedback is suppressed when the label is hidden — it requires a
+    // visible label to provide context. Supporting text and the counter are
+    // independent of label visibility and remain shown.
+    const showFeedback =
```
Chocolate-lightning: `invalidFeedback` is rendered only when `hasVisibleLabel` is true. A field with `hideLabel` and an error therefore has an invalid border/icon but no error message and no feedback ID in aria-describedby. Should error messaging be conditional on `hideLabel`?

components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,304 @@
...
+      rows = 3,
+      maxLength,
+      value,
+      defaultValue,
+      onChange,
+      onPointerDown,
+      onFocus,
+      onBlur,
+      'aria-label': ariaLabelProp,
+      ...textareaProps
```
Chocolate-lightning, line 90: A consumer-provided `aria-describedby` may overwrite the generated IDs since it is not being destructured.

components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,304 @@
...
+          const remaining = maxLength - newLength;
+          const prev = prevRemainingRef.current;
+          prevRemainingRef.current = remaining;
+
+          // Announce only at meaningful milestones (ZeroHeight: throttle
+          // announcements — not on every keystroke).
+          if (prev !== null) {
+            if (remaining < 0) {
+              setAnnouncement(
+                `${Math.abs(remaining)} over the ${maxLength} character limit`,
```
Chocolate-lightning: We may need to accept translated counter/accessibility strings or formatter callbacks as props so we don't have english only text within a component. This breaks our i18n requirements currently.

components/textarea/textarea.css
```diff
@@ -0,0 +1,243 @@
...
+  border: var(--mds-stroke-weight-sm) solid
+    var(--mds-border-interactive-secondary-default);
+  border-radius: var(--mds-border-radius-sm);
+
+  /* Animate default -> hover fill; border state changes remain immediate. */
+  transition: background-color 150ms ease;
+}
+
+.mds-textarea[class] .mds-textarea-field--no-resize[class] {
+  resize: none;
```
Chocolate-lightning, line 97: ZeroHeight defines read-only as populated, selectable content and states that the resizer still renders when `resizable` is true. So this should only be needed when `resizable` is false.

components/textarea/textarea.css
```diff
@@ -0,0 +1,243 @@
...
+  color: var(--mds-text-subtle);
+}
+
+.mds-textarea[class] .mds-textarea-required {
+  color: var(--mds-text-danger);
+  font-size: var(--mds-font-size-paragraph-small);
+  font-weight: var(--mds-font-weight-medium);
+}
+
+/* Info icon-button beside the label — reuses Button ghost sm for the hit area,
```
Chocolate-lightning, line 49: The info button's padding is explicitly zeroed and the button is constrained to the 16px label-row height. The MSE design checklist requires 44 x 44px hit areas, and ZeroHeight describes padding around the icon hit area. The icon may be visually aligned, but the interactive target may currently be too small for touch and motor accessibility.

components/textarea/textarea.css
```diff
@@ -0,0 +1,243 @@
+/* ==========================================================================
```
Chocolate-lightning, line 1: I see various hard definitions of geometry & transitions for example, are we happy with this? Might need to check with Boris.

components/textarea/Textarea.stories.tsx
```diff
@@ -0,0 +1,311 @@
...
+      description:
+        'Pre-translated error message shown below the field when invalid is true.',
+      control: { type: 'text' },
+      if: { arg: 'invalid', truthy: true },
+    },
+    supportingText: {
+      description: 'Helper text shown below the field in non-error state.',
+      control: { type: 'text' },
+    },
+    showCounter: {
```
Chocolate-lightning, line 62: `showCounter` is conditionally displayed with `if: { arg: 'maxLength', truthy: true }`. A `maxLength` of `0` hides the control even though it is a provided value, and the number control has no explicit reset/validity guidance.

components/textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,335 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { createRef } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+import { Textarea } from './Textarea';
+
+describe('Textarea: Unit Test', () => {
```
Chocolate-lightning, line 7: The passing unit tests cover the main states, but not the hidden-label error case, consumer description merging, pointer-vs-keyboard focus, read-only resize behaviour, disabled content guidance, controlled values above maxLength, or milestone announcement text. The existing * required assertion also queries the document globally rather than proving the indicator belongs to this label.

components/textarea/textarea.css
Chocolate-lightning, line 1: Has this component been tested with form element siblings to ensure the padding and indentation is correct? Would be good to add it to the MSE with a screenshot or something. wdyt?

components/textarea/Textarea.figma.tsx
```diff
@@ -0,0 +1,125 @@
+import figma from '@figma/code-connect';
+import { Textarea } from './Textarea';
+
+const url =
+  'https://www.figma.com/design/6u22h9GzDI0wWL5yHwQ1jx/Text-area?node-id=13934-4878';
```
Chocolate-lightning: This should likely be updated to reference bPRkRtSszcbWw9f9p9rXvA / the main Figma file.

components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,304 @@
...
+      invalid,
+      invalidFeedback,
+      supportingText,
+      resizable = true,
+      showCounter = false,
+      infoTooltipLabel,
+      className,
+      id: idProp,
+      required,
+      disabled,
```
Chocolate-lightning, line 78: ZeroHeight says disabled should be used for unavailable input and read-only for content that remains readable/copyable. The MSE checklist additionally says a disabled textarea needs an explanation. The current disabled story includes supporting text, but the component does not enforce or document that expectation for consumers.

Chocolate-lightning (changes requested):

components/textarea/Textarea.tsx
andrewgos: confirmed with boris, it was a miscommunication on ours

im making the invalidFeedback to always appear regardless hideLabel status (just like the supportingText)

components/textarea/Textarea.tsx
andrewgos: `aria-describedby` is now destructured and used along with the IDs

components/textarea/textarea.css
andrewgos: updated. now read-only with `resizable:true` will have the resizer

components/textarea/Textarea.tsx
andrewgos: added props that takes in either string or formatter for these announcements

components/textarea/textarea.css
andrewgos: @borisbasque i'll need your opinion on this. any thoughts about the 'i' button very small for accessibility purposes?

components/textarea/textarea.css
andrewgos: @Chocolate-lightning thanks for raising your concern, can you give us some examples of the hard definitions?

components/textarea/Textarea.stories.tsx
andrewgos: good pickup.

making the condition to be `exists` instead of `truthy`

components/textarea/Textarea.test.tsx
andrewgos: unit tests updated with more tests

components/textarea/Textarea.tsx
andrewgos: Props updated with descriptions 👍

components/textarea/Textarea.figma.tsx
andrewgos: ah nice catch! url updated

components/textarea/textarea.css
andrewgos: Tested along with checkbox+radio+choicebox, things looks aligned and well 👍 

Screenshot 2026-08-27 at 1 28 30 pm

components/textarea/textarea.css
borisbasque: arguably not a hit area because it's triggered on hover. we can't get to 44px in any way

components/textarea/textarea.css
borisbasque: we don't have tokens for transitions yet unfortunately but it's something we will need to revisit in future. As for sizes it's ok to hardcode some things

Chocolate-lightning: Yep, figured as much. Thanks for weighing in @borisbasque

components/textarea/textarea.css
Chocolate-lightning: Alright, thanks for the input @borisbasque I wanted to double check since that is what the MSE had in it.

components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,371 @@
...
+      'aria-describedby': ariaDescribedByProp,
+      ...textareaProps
+    }: TextareaProps,
+    ref,
+  ) => {
+    const generatedId = useId();
+    const id = idProp ?? generatedId;
+    const isInvalid = !!invalid;
+    const hasVisibleLabel = !hideLabel;
+    const ariaLabel = hideLabel ? (ariaLabelProp ?? label) : undefined;
```
Chocolate-lightning, line 98: Maybe we need a small dev warning that when an aria-label is passed with a visible label that the aria-label will be ignored in lieu of the provided label. wdyt?

components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,371 @@
...
+          )}
+        </div>
+        {(footerText || showCounterEl) && (
+          <div className="mds-textarea-footer">
+            {feedbackId ? (
+              <span
+                id={feedbackId}
+                className={
+                  showFeedback
+                    ? 'mds-textarea-feedback invalid-feedback'
```
Chocolate-lightning: I think with the invalid-feedback class, whilst inline with our requirement to have BS classes where applicable is not firing here correctly since the class is applied via sibling elements and considering the nesting of the textarea and this footer section, they are not siblings. Maybe we could either remove it or add a comment mentioning this issue? wdyt?

components/textarea/Textarea.figma.tsx
```diff
@@ -0,0 +1,125 @@
+import figma from '@figma/code-connect';
+import { Textarea } from './Textarea';
+
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/Moodle-Design-System?node-id=13934-4878';
+
+const baseProps = {
```
Chocolate-lightning, line 7: We should probably include a CC example of the info button inclusion in this file.

components/textarea/textarea.css
```diff
@@ -0,0 +1,250 @@
...
+
+/* Pointer-initiated focus (mouse/touch): suppress the ring — the primary border
+   signals the selected/typing state. The ring is reserved for keyboard navigation. */
+.mds-textarea[class] .mds-textarea-field--pointer-focus[class]:focus-visible {
+  outline: none;
+  border-color: var(--mds-border-interactive-primary-default);
+}
+
+/* Invalid */
+.mds-textarea[class] .mds-textarea-field.is-invalid[class] {
```
Chocolate-lightning, line 141: Is the combination state of disabled and invalid a supported use case? If so the invalid border color is being overridden with the disabled border color. Maybe something to confirm.

components/textarea/Textarea.tsx
```diff
@@ -0,0 +1,371 @@
...
+            readOnly={readOnly}
+            required={required}
+            rows={rows}
+            maxLength={maxLength}
+            value={value}
+            defaultValue={defaultValue}
+            aria-invalid={isInvalid ? true : undefined}
+            aria-label={ariaLabel}
+            aria-describedby={describedBy}
+            aria-readonly={readOnly ? true : undefined}
```
Chocolate-lightning: The native `readonly` attribute already exposes the correct accessible state automatically. This explicit aria-readonly is harmless but unnecessary duplication so it could be safely removed.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Great work on this so far, only a couple of small items left to look at before this is ready.

Also make sure to check out the "React Hook useEffect has a missing dependency" notice in the PR as it flags a warning when running tests.

Cheers,

components/textarea/Textarea.tsx
andrewgos: fair call. adding it in

components/textarea/Textarea.figma.tsx
andrewgos: good catch. it wasnt initially obvious with the info button not visible in the figma node. adding it in

components/textarea/Textarea.stories.tsx
```diff
@@ -0,0 +1,391 @@
...
+      if: { arg: 'hideLabel', eq: false },
+    },
+    showCounter: {
+      description:
+        'Shows a live character counter. Requires maxLength to be set.',
+      control: { type: 'boolean' },
+      table: { defaultValue: { summary: 'false' } },
+      if: { arg: 'maxLength', exists: true },
+    },
+    counterAriaLabel: {
```
Chocolate-lightning, line 69: Is there a reason we are disabling control for this property and the other Announcement props?

Chocolate-lightning (changes requested):

components/textarea/Textarea.tsx
andrewgos: fair call... it was mainly to keep it consistent with checkbox and radio. but i agree, lets clean it up for the textarea.

Removed the use of `invalid-feedback`, while also renaming some existing ones to be clearer with its intention to be `.mds-textarea-invalid-feedback` and `showInvalidFeedback`

components/textarea/textarea.css
andrewgos: good catch, just like checkbox and radio, disabled state overrides the invalid state. you spotted a bug where the invalid icon still appear there. im fixing that.

components/textarea/Textarea.tsx
andrewgos: removed 👍

components/textarea/Textarea.stories.tsx
andrewgos: it was mainly to avoid confusion with the fact that it could take in a formatter function, but it wasn't a biggie, i guess it's fine to open up a text control 👍  adding it in

Chocolate-lightning (approved):
Kia ora @andrewgos,

Thanks for actioning the latest round of feedback. I think this is now in a state to merge in.

Looking at the tests locally & within my LMS PVT instance, It is looking good and the props work as expected and the functionality is working as expected.

Accepting this.

components/textarea/textarea.css
andrewgos: added into the MSE that this is approved as exception

> Info ("i") button hit area — exception approved Hit area is smaller than the standard 44×44px minimum. Approved exception: the icon is hover-triggered, not a tap/click target, so the touch-target guideline doesn't apply here.

PR #329, opened 2026-09-01
feat(Input, PasswordInput): Add new components

andrewgos:
## Description

Adds two new form field components: `Input`, a single-line text-like field (email/number/tel/text/url) with label, supporting text, and validation feedback; and `PasswordInput`, built on the same shared `BaseInput` primitive, adding a fixed lock icon and a show/hide visibility toggle. Both components share layout, validation, and styling logic through an internal `BaseInput`.

## Key implementation details

- **Shared `BaseInput` primitive** — `Input` and `PasswordInput` both compose an internal, unexported `BaseInput` that owns label rendering, supporting/feedback text, invalid-state styling, and start-icon/trailing-action slots. `Input` locks `inputType` to text-like values; `PasswordInput` locks `startIcon` to a fixed lock icon and `trailingAction` to the visibility-toggle button, so those props aren't user-overridable on either public component (enforced via `Omit` on `PasswordInputProps`).
- **Native HTML validation is captured, not just accepted** — `BaseInput` listens for the native `onInvalid` event, calls `preventDefault()` to suppress the browser's default validation bubble, and stores `validationMessage` in state so it can render through the same `invalidFeedback`/`mds-input-feedback` UI as consumer-supplied errors. It self-clears on the next `onChange` once `checkValidity()` passes. `effectiveInvalid` is `invalid || !!nativeMessage`, so native failures (`required`, type mismatch, etc.) mark the field invalid even if the consumer never passes `invalid`.
- **Pointer-vs-keyboard focus tracking** — a `isPointerDownRef` + `isPointerFocus` state pair distinguishes a pointer-initiated focus from a keyboard one, toggling `mds-input-field--pointer-focus` so `:focus-visible` styling (focus ring) is suppressed for mouse clicks but preserved for keyboard navigation.
- **Runtime icon guard** — `startIcon` is validated at runtime via `isValidElement` + tag-name check (`<i>` or `<svg>` only); anything else is dropped and logged via `console.error` in dev rather than rendered, since arbitrary elements would break the fixed icon sizing/positioning contract.
- **Type restriction with graceful fallback** — `Input`'s `type` prop is constrained to `email | number | tel | text | url` (password intentionally excluded — that's what `PasswordInput` is for); an invalid value falls back to `text` with a dev-only `console.warn`.
- **PasswordInput composes `Button`** for the visibility toggle (`variant="ghost"`) rather than a bare `<button>`, so it inherits the design system's button a11y/interaction behavior; the toggle's `aria-label` swaps between `passwordToggleShowLabel`/`passwordToggleHideLabel` based on visibility state, and toggling flips `inputType` between `password`/`text` rather than using a separate reveal mechanism.
- **CSS padding is computed, not fixed** — `mds-input-field--with-start-icon`, `--with-status`, and `--with-action` each reserve inline padding via `calc()` against `--mds-spacing-*`/`--mds-icons-*` tokens, including a combined case where both a status icon and a trailing action render together, to keep icon/text clearance in sync with the token scale instead of hardcoded pixel offsets.
- **`::-ms-reveal`/`::-ms-clear` suppressed** in `password-input.css` so legacy Edge/IE's built-in password affordances don't visually double up with the custom toggle.

https://moodle.atlassian.net/browse/MDS-640

## Screenshots/Previews

Screenshot 2026-09-01 at 3 19 24 pm

Snapshots can be found in Chromatic

- [x] All 5 required files present: `ComponentName.tsx`, `component-name.css`, `ComponentName.test.tsx`, `ComponentName.stories.tsx`, `index.tsx` (Input and PasswordInput each have their own `.tsx`/`.css`/`.test.tsx`/`.stories.tsx`; both share the folder's single `index.tsx` and internal `BaseInput.tsx`)
- [x] `ComponentName.figma.tsx` Code Connect file created and published — `Input.figma.tsx` and `PasswordInput.figma.tsx` added; confirm publish status in ZeroHeight/Figma before merge

### TypeScript and props

- [x] Props interface extends the correct React HTML element interface (e.g. `ButtonHTMLAttributes<HTMLButtonElement>`) — `InputHTMLAttributes<HTMLInputElement>`
- [x] JSDoc comment on each prop — present on `InputProps` and the two `PasswordInputProps` additions; internal `BaseInputProps` (not directly exported to consumers) is undocumented, which is acceptable for a shared internal primitive
- [x] Constrained props use `allowedValues` runtime validation with a graceful fallback to a documented default — `Input`'s `type` falls back to `text` with a dev warning

### Internationalisation

- [x] All user-facing strings accepted as props — no hardcoded text in JSX (`label`, `invalidFeedback`, `supportingText`, `passwordToggleShowLabel`, `passwordToggleHideLabel`, etc.)
- [x] RTL story added if the component has directional layout — present in both `Input.stories.tsx` and `PasswordInput.stories.tsx`

### CSS and tokens

### Tests and stories

- [x] Accessibility tested — Axe WCAG AA scan runs automatically via CI on all stories tagged `test` — stories tagged `test`/`stable`

### Breaking changes

- [x] Breaking change assessment complete — no prop, export, or token removed/renamed without a major version bump (additive only)

### AI context

- [x] Instruction files updated if this component introduces new patterns or anti-patterns — `component-index.instructions.md` updated with `Input` and `PasswordInput` entries and doc links

## Cross-system parity

- [x] Component name, prop names, and variant names are consistent across code, Storybook, Figma, and Zeroheight — please verify against the Figma/ZeroHeight source before merge

## Additional Notes

`BaseInput` is an internal, unexported-from-index shared primitive (label row, feedback/supporting text, focus/pointer-focus handling, native validation capture) that both `Input` and `PasswordInput` compose over — it intentionally has no standalone stories/tests since it isn't a published component in its own right; coverage lives in the `Input` and `PasswordInput` test/story files instead.


components/input/BaseInput.tsx
```diff
@@ -0,0 +1,262 @@
...
+    // when the consumer marks the field invalid without supplying invalidFeedback text.
+    const [nativeMessage, setNativeMessage] = useState<string | undefined>(
+      undefined,
+    );
+
+    // Native validation failures (e.g. required, type mismatch) mark the field invalid
+    // even if the consumer never passed the invalid prop.
+    const effectiveInvalid = invalid || !!nativeMessage;
+
+    const showInvalidFeedback =
```
Chocolate-lightning, line 88: We need some readOnly checks for this section of logic since the ZH docs state that "Disabled and Read-only don't carry an invalid state" and documents Invalid as a modifier that only layers onto Default/Hover/Active/Focus. So this means we are breaking from the documented cases currently and need to add some guardrails.

components/input/input.css
```diff
@@ -0,0 +1,223 @@
+/* Input component
```
Chocolate-lightning, line 1: Once again we are hitting a MSE error related to 40x40px button hit areas. The current implementation should satisfy WCAG 2.2 SC 2.5.8 Target Size (Minimum, AA), which sets the bar at 24×24px. It seems that the MSE should be updated and we need to check if any others have this _stricter_ 40x40px requirement.

components/input/password-input.css
```diff
@@ -0,0 +1,16 @@
+/* PasswordInput component
```
Chocolate-lightning, line 1: The masked password i.e. `···········`  should use a bold weight token rather than using the normal weight as that is purely for shown/unmasked passwords.

components/input/Input.test.tsx
```diff
@@ -0,0 +1,150 @@
+import { render, screen } from '@testing-library/react';
```
Chocolate-lightning: We should add a couple more unit tests just to cover a couple of baseinput pieces of functionality:
1. Fire an invalid/blur sequence and assert aria-invalid/feedback text appear without invalidFeedback supplied, then clear on a valid onChange
2. Simulate pointerdown → focus and assert mds-input-field--pointer-focus is applied, vs. a plain focus (keyboard) not applying it

components/input/PasswordInput.tsx
```diff
@@ -0,0 +1,82 @@
...
+      trailingAction={toggleButton}
+      className={classes.join(' ')}
+      disabled={disabled}
+      {...props}
+    />
+  );
+}
+
+export const PasswordInput = forwardRef<HTMLInputElement, PasswordInputProps>(
+  PasswordInputImpl,
```
Chocolate-lightning: I am curious if there was a reason behind this since the standard Input component does not split out invocation & rendering portions of the component call. We could maybe look at removing the abstraction or at-least documenting why it has to be this way. wdyt?

components/input/password-input.css
Chocolate-lightning, line 1: Adding a small note to cover what has already been discussed, the mask / unmask icons should be replaced with SVG files to prevent them from overflowing from their 12x12px constraints like they currently due which makes the icons look off center within the button.

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Overall this is a solid implementation with just a couple of minor nitpicks basically.

They should be simple as to resolve.

Cheers,

components/input/password-input.css
andrewgos: replaced the FA icons with SVG

components/input/BaseInput.tsx
andrewgos: such a great pickup.... its happening in textarea too

 invalid=true with readonly=true showing readonly with red lines which we do not want.

i will fix it for Input and Textarea

components/input/password-input.css
andrewgos: In Figma I can see the `Style` is marked as `bold`, but `font-weight` token is still `regular`. can you confirm @borisbasque ?
Screenshot 2026-09-04 at 4 17 47 pm

components/input/PasswordInput.tsx
andrewgos: agree, removing it. i guess it was a leftover while re-arranging the BaseInput structuring.

components/input/Input.test.tsx
andrewgos: tests added

components/input/input.css
andrewgos: I suppose you are referring to the FieldInfoButton just like the textarea?

I've added the note to the MSE

> Info ("i") button hit area — exception approved Hit area is smaller than the standard 44×44px minimum. Approved exception: the icon is hover-triggered, not a tap/click target, so the touch-target guideline doesn't apply here.

components/input/password-input.css
borisbasque: resolved

Chocolate-lightning: Decision: bold applied only to the masked input.

Chocolate-lightning (approved):
Kia ora @andrewgos,

Thanks for taking my feedback onboard. After testing this in Storybook, within Moodle LMS and running local tests, I think this is ready to go. I will note that the MSE needs to be completed for this component and ZH will likely need a once over as well.

The patch and general implementation is straightforward and our usual CI appears to be happy with the arrangement as well.

Accepting this in,

PR #338, opened 2026-09-11
feat(SearchInput): Add new SearchInput component

andrewgos:
## Description

Adds a new `SearchInput` component built on top of `BaseInput`. It always renders a decorative magnifying-glass icon at the start of the field and shows a `CloseButton` clear affordance once the field has a value, wiring the native input's value setter so clearing works correctly for both controlled and uncontrolled usage. `type="search"` is fixed on the underlying input and cannot be overridden.

https://moodle.atlassian.net/browse/MDS-642

- [x] All 5 required files present: `SearchInput.tsx`, `search-input.css`, `SearchInput.test.tsx`, `SearchInput.stories.tsx`, `index.tsx`
- [x] `SearchInput.figma.tsx` Code Connect file created and published

### TypeScript and props

- [x] Props interface extends the correct React HTML element interface (via `BaseInputProps`, itself extending `InputHTMLAttributes<HTMLInputElement>`)

### Internationalisation

- [ ] RTL story added if the component has directional layout

### CSS and tokens

### Tests and stories

### Breaking changes

### AI context

## Cross-system parity

## Additional Notes

Badges and the multiselect chevron (Figma "Badges=Yes" variant) are out of scope for this PR — that content models a searchable multiselect combobox, a separate component from this single-value `SearchInput`, and may be tracked as follow-up work.


components/input/SearchInput.figma.tsx
```diff
@@ -0,0 +1,43 @@
+import figma from '@figma/code-connect';
+import { SearchInput } from './SearchInput';
+
+const url =
+  'https://www.figma.com/design/bPRkRtSszcbWw9f9p9rXvA/branch/CZFqyV4YnE4ZfeQ2yuZ5IS/Moodle-Design-System?node-id=15029-1542';
```
andrewgos, line 5: waiting for the figma branch to be merged before updating this link to point to the figma main branch

components/input/SearchInput.figma.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+// The "Badges=Yes" variant (tag/chip content with a chevron) models a
+// searchable multiselect combobox, not this single-value SearchInput — only
+// the "Badges=No" variants are mapped here. See the generation report for
+// details.
+
+figma.connect(SearchInput, url, {
+  variant: { Content: 'Empty', Badges: 'No' },
```
Chocolate-lightning: Each of the connect() calls include Badges which is not surfaced in the Figma designs but Tags instead, this will likely need an update.

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+    }
+  };
+}
+
+export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
+  function SearchInput(
+    {
+      className,
+      disabled,
+      readOnly,
```
Chocolate-lightning: The ZeroHeight docs state: "Read-only doesn't exist for Search. A locked, non-removable value is handled at the individual badge level, not as a state on the whole field." 

So we should reflect that in the code as well. This will need some work across the component to remove but should not prove too difficult.

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export interface SearchInputProps extends Omit<
+  BaseInputProps,
+  'inputType' | 'startIcon' | 'trailingAction' | 'type'
+> {
+  /** Accessible label for the button that clears the field's current value. */
+  clearLabel: string;
+}
+
+// Fixed per design: search fields always show a magnifying-glass icon at the start of the field.
```
Chocolate-lightning, line 40: Figma exposes Leading icon as a `BOOLEAN` component property. ZeroHeight states: "Leading icon (optional) — fixed to a magnifier glyph when shown", and the Interaction behaviour section adds a case where it is actively hidden: "When the search input is filled with badges, the leading icons disappear." 

Is this a holdover or outcome after deciding that this component is purely for singular search input rather than including the tags type search?

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+      ...props
+    },
+    ref,
+  ) {
+    const inputRef = useRef<HTMLInputElement | null>(null);
+    const isControlled = value !== undefined;
+    const [uncontrolledHasValue, setUncontrolledHasValue] = useState(() =>
+      Boolean(defaultValue),
+    );
+    const hasValue = isControlled ? Boolean(value) : uncontrolledHasValue;
```
Chocolate-lightning, line 89: We should probably trim whitespace to prevent cases like ' ' being treated as a valid input value. Further to this, MSE §18 requires "Empty submit is prevented (whitespace-only counts as empty)". But again, this may be more important when the Search with tags is made. Do we have a followup or MDS for that yet?

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+const searchIcon = (
+  <i
+    className="fa-solid fa-magnifying-glass mds-input-search-icon"
+    aria-hidden="true"
+  />
+);
+
+// Merges the ref this component needs (to imperatively clear the native
+// input's value) with whatever ref the consumer forwarded in.
+function mergeRefs<T>(...refs: Array<Ref<T> | undefined>) {
```
Chocolate-lightning, line 50: Considering that search inputs typically go hand in hand with debounce functions, we would likely be calling this function constantly and passing back null references on any passed callbacks. We should likely do something along the lines of: `const setRefs = useCallback(mergeRefs(ref, inputRef), [ref]);`

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
+import {
```
Chocolate-lightning, line 1: The ZeroHeight docs need a proofreading pass, and give Empty vs Placeholder genuinely different descriptions (presumably: Empty = no placeholder text configured; Placeholder = placeholder text shown)

components/input/SearchInput.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+    readOnly: false,
+    clearLabel: 'Clear search',
+  },
+  argTypes: {
+    label: {
+      description: 'Visible label text above the search field.',
+      control: { type: 'text' },
+      table: { type: { summary: 'string' } },
+    },
+    placeholder: {
```
Chocolate-lightning, line 40: Placeholder is documented with no optionality signal and no defaultValue in its table - a reader cannot tell it is optional, and ZeroHeight has substantive guidance for it ("hint at what can be searched (e.g. 'Search courses') … never as a substitute for the label")

components/input/SearchInput.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+      table: { type: { summary: 'string' } },
+      if: { arg: 'hideLabel', eq: false },
+    },
+  },
+} satisfies Meta<typeof SearchInput>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {};
```
Chocolate-lightning, line 161: There is no guidance anywhere on attaching a debounced handler, despite MSE §2 flagging the absence of a shared debounce as a Must Have failure in the current implementation ("Nothing defines how fast instant search fires")

components/input/SearchInput.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+    hideLabel: {
+      description:
+        'When true, hides the visible label. Provide aria-label or label for accessible naming.',
+      control: { type: 'boolean' },
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    invalid: {
```
Chocolate-lightning, line 55: `Invalid`'s description restates the prop name and omits the one rule that actually governs it. ZeroHeight states: "Invalid state applies only when content is Empty … Filled or badge states don't carry an invalid treatment." Note the Invalid story contradicts that rule directly, passing `defaultValue: 'xyz123' alongside invalid: true`

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+      } else if (ref) {
+        // RefObject.current is technically read-only from the consumer's
+        // perspective, but forwardRef/useRef both hand back a mutable object.
+        (ref as { current: T | null }).current = node;
+      }
+    }
+  };
+}
+
+export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
```
Chocolate-lightning, line 64: MSE §1 lists "Keyboard interaction tested — Tab through field, submit and clear controls; Enter submits; Escape clears" as a Dev Must Have. No Escape handling exists. Complicating this: `search-input.css` hides `::-webkit-search-cancel-button`, and browsers' native Escape-to-clear on `type="search"` would in any case desync the shadow state

components/input/SearchInput.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+    readOnly: {
+      description:
+        'Use when existing content should stay readable/copyable while editing is blocked.',
+      control: { type: 'boolean' },
+      table: {
+        type: { summary: 'true | false' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    clearLabel: {
```
Chocolate-lightning, line 82: Considering this is a required prop, it is quite low in the prop list in Storybook and kind of gets buried. It'd be nice to surface it a bit earlier if possible.

components/input/SearchInput.tsx
Chocolate-lightning, line 1: MSE §14 specifies "Component lives at components/search-input/". It lives at components/input/ alongside Input, PasswordInput and the shared BaseInput. We should update the MSE to better reflect where it will live instead otherwise it would be treated as a defect.

components/input/SearchInput.test.tsx
```diff
@@ -0,0 +1,179 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { useState } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+import { SearchInput } from './SearchInput';
+
+describe('SearchInput: Unit Test', () => {
```
Chocolate-lightning, line 7: Specific tests worth having: the whitespace-only case; the imperative-clear desync; that the control reports as `searchbox` not `textbox` MSE §18 asks for this by name — "Add one test asserting the control reports itself as a searchbox and not as a plain text field — that's the failure mode of borrowing Text Input's surface" — and getByRole('searchbox') would assert it directly, where the current test only checks the type attribute); clear-then-type-again, which exercises the state round-trip; and the object-ref branch.

components/input/SearchInput.stories.tsx
Chocolate-lightning, line 161: Hover and Focus pseudo-state stories would give Chromatic something to diff if we were to add a small matrix story.

components/input/SearchInput.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+import { SearchInput } from './SearchInput';
+
+const meta = {
+  title: 'Components/Input/SearchInput',
+  component: SearchInput,
+  parameters: {
+    layout: 'centered',
+    a11y: {
+      options: {
+        rules: {
```
Chocolate-lightning: This a11y rule is switched off at the meta level, so it silences every story in the file, not just the ones with the clear button. WithInfoButton didn't need it — the shared FieldInfoButton is actually 12×12px with a 14px glyph inside it, which is a genuine target-size failure that's got nothing to do with search. Can we scope the suppression to just the story it's meant for, and raise the info button as its own issue against FieldInfoButton?

components/input/SearchInput.tsx
Chocolate-lightning, line 89: `hasValue` is tracked in React state that only updates via `onChange`, so anything that sets the input's value outside that path, the native setter, autofill, a consumer holding the forwarded ref would leave the clear button showing on an empty field. Worth deriving `hasValue` from the DOM (or the native input event) instead of shadowing it, especially since `handleClear` itself relies on that same native-setter.

components/input/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+        onClick={handleClear}
+      />
+    ) : undefined;
+
+    return (
+      <BaseInput
+        ref={mergeRefs(ref, inputRef)}
+        inputType="search"
+        startIcon={searchIcon}
+        trailingAction={clearButton}
```
Chocolate-lightning, line 208: The clear button is absolutely positioned over the input rather than laid out in flow, so the input's own hit box still extends underneath it. Might be worth restructuring `.mds-input-field-wrapper` so the trailing action is a real flex sibling instead of an overlay; that'll touch Input/PasswordInput/Textarea too, so probably wants its own PR. wdyt?

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Good work on this component so far, I spotted a few items I'd like to see addressed and a couple of questions to go through as well.

If anything is unclear, just reach out.

components/input/SearchInput.figma.tsx
andrewgos: thanks for noting that. updated to match the latest figma

components/input/SearchInput.tsx
andrewgos: no worries. updated across the component/tests/stories 👍

components/input/SearchInput.tsx
andrewgos: yes the scope was intentional made without Tags/ComboBox. For now we are keeping the leading icons to always appear, we can the put the conditional appearance when we comes with the Tags and ComboBox later.

components/input/SearchInput.tsx
andrewgos: That would be the scope for Combobox component https://moodle.atlassian.net/browse/MDS-643

components/input/SearchInput.tsx
andrewgos: nice one. adding it in

components/input/SearchInput.tsx
andrewgos: cc @edumoodle

components/input/SearchInput.stories.tsx
andrewgos: good point, editing the description in storybook to be more explicit of it being optional

components/input/SearchInput.stories.tsx
andrewgos: fair call for the use of search input. adding the debounced handler in.

components/input/SearchInput.stories.tsx
andrewgos: @edumoodle @borisbasque

Do we want this to be part of the logic? ie. remove the invalid look when value is filled (even though invalid=true) 

Or just part of documentation only?

components/input/SearchInput.tsx
andrewgos: esc-to-clear feature added. along with the storybook (excluded from chromatic)

components/input/SearchInput.stories.tsx
borisbasque: if it's filled it can't be invalid

components/input/SearchInput.stories.tsx
andrewgos: I'm afraid it's a technical limitation due to the nature of our code being extended from BaseInput. The ordering is pretty rigid with what came from the autodoc.

Other options:
- We could use the feature "sort:requiredFirst", but it will affect the ordering of everything else as alphabetical
- Another way is to utilise the "Category" feature of Storybook to split out some props to another section; however, it'll be something new introduced while the other components didn't do this.

I'm leaning towards leaving just this one prop as is, compared to impacting the other props What do you think? 🤔

components/input/SearchInput.tsx
andrewgos: updated the MSE 👍

components/input/SearchInput.test.tsx
andrewgos: tests updated with empty-state cases

components/input/SearchInput.stories.tsx
andrewgos: good call. adding it all to Input / PasswordInput / SearchInput

components/input/SearchInput.stories.tsx
andrewgos: nice spotting. narrowing it to only stories with FieldInfoButton.

components/input/SearchInput.tsx
andrewgos: added active update on the value using `useEffect`

components/input/SearchInput.tsx
andrewgos: totally valid. this was something that was considered while fixing the "CloseButton hit area is too small and too close to the invalid icon" issue.

maybe bringing this in will bring a proper solution to the ignored a11y error `'target-size': { enabled: false },` too.

Im working on it in this PR.

components/input/SearchInput.stories.tsx
andrewgos: thanks for confirming. im making the invalid state/look to be automatically supressed whenever the field is filled (even though the consumer do invalid:true)

this also fixes the problem with the hit area of close button being too close to the invalid icon = because it won't happen again with this logic! so i'm cleaning up the ignores rules of a11y too.

components/input/SearchInput.tsx
Chocolate-lightning: Fair call but I think since we are determining display logic i.e. the close button, we should likely trim the whitespace only values unless we can configure how to handle that display logic by the invocation i.e. combobox trims whitespace so it calls the search input with that logic to handle it. wdyt?

components/input/SearchInput.stories.tsx
Chocolate-lightning: Yep that is fine, wanted to flag it in case we were able to control it to a certain extent.

.github/instructions/component-index.instructions.md
```diff
@@ -27,6 +27,7 @@ Quick reference for AI agents and developers.
 | Pagination      | Page navigation control                                                                                       | totalPages, currentPage, onPageChange, ariaLabel                                                  |
 | PasswordInput   | Single-line password input built on Input, with a lock icon and show/hide visibility toggle                   | label, invalid, invalidFeedback, supportingText, passwordToggleShowLabel, passwordToggleHideLabel |
 | ProgressBar     | Visual progress indicator with status and label variants                                                      | value, min, max, status, labelVariant, title                                                      |
+| SearchInput     | Single-line search input built on Input, with a magnifying-glass icon and a clear-value button                | label, invalid, invalidFeedback, supportingText, clearLabel                                       |
```
Chocolate-lightning: We likely need to add the debounce props here no?

Chocolate-lightning (changes requested):
Kia ora @andrewgos,

Thanks for actioning my feedback, I have small request left as well as what I mentioned in chat.

Cheers,

.github/instructions/component-index.instructions.md
andrewgos: debounce prop added 👍

andrewgos:
✅ Fixed the native validation `pattern` that bubbles up for the SearchInput.
✅ Added the debounceProp to the component index file

Ready for another look @Chocolate-lightning

components/input/SearchInput.tsx
andrewgos: sorry missed this unresolved comment.

thats a fair point, we can definitely implement it on this level without worrying who's invocating

adding the trims on the `hasValue` check (and several other related places) as well as the `onDebouncedChange` callback

components/input/SearchInput.figma.tsx
andrewgos: figma branch updated

andrewgos:
✅ Updated figma branch
✅ Resolved the comment around the trimming white spaces
✅ Commits squashed

andrewgos:
✅ Resolved the issue https://moodle.atlassian.net/browse/MDS-675 in this PR (that is to swap `min-height: 40px` to `height: 42px`

